PR #4, opened 2019-12-09
Update 01_foundation.ipynb

radekosmulski:
I am finding `BypassNewMeta` really hard to follow and to understand it's use case. Thought I'd maybe raise this as `fastcore` is now a standalone component but please disregard if this is not something that makes sense giving your time to.

I think the `x=None` in signature might be misleading, AFAIK it can never work when passed `None` (or not passed anything). If that is the case removing this adds a little bit of legibility.

### DIFF
--- a/nbs/01_foundation.ipynb
+++ b/nbs/01_foundation.ipynb
@@ -208,7 +208,7 @@
     "#export\n",
     "class BypassNewMeta(FixSigMeta):\n",
     "    \"Metaclass: casts `x` to this class if it's of type `cls._bypass_type`, initializing with `_new_meta` if available\"\n",
-    "    def __call__(cls, x=None, *args, **kwargs):\n",
+    "    def __call__(cls, x, *args, **kwargs):\n",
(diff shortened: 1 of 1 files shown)


radekosmulski:
I understand the functionality better now. Problem is I don't have a suggestion on how this could be refactored (if it could be at all retaining the conciseness).

What confused me was that the description reads:
```
"Metaclass: casts `x` to this class if it's of type `cls._bypass_type`, initializing with `_new_meta` if available"
```

But the functionality is:
image

The `horse` got casted to a `Unicorn` despite `_bypass_type` not being specified.

But then I realized that if no `_bypass_type` is specified it defaults to object. Everything makes sense now.

radekosmulski:
I think I figured out how this works :)

image

radekosmulski:
As far as I can tell, the `_new_meta` of `BypassNewMeta` is not used anywhere. If that is the case, it would be great to remove this from the code.

sgugger:
Closing the PR as you seem to have made sense of how things work, please reopen if I'm mistaken.
For `_new_meta` it may very well be that we don't need it anymore, feel free to suggest a PR removing it and we can test if it breaks things.

PR #5, opened 2019-12-27
test_fail for functions with arguments

brunosan:
Allows to test functions with arguments.

Also added these examples so they appear on the documentation:
```python
def _fail(a): return
test_fail(_fail)
test_fail(test_fail,_fail,0)
```


brunosan:
Just realized this is equivalent to using `test_fail(lambda:  f(a,b))`.

PR #6, opened 2020-01-27
Fix nested delegation

rbracco:
Removes two lines from delegates in foundations nb. `if hasattr(from_f,'__delwrap__'): return f` and `from_f.__delwrap__ = to_` were of unknown purpose and were breaking nested delegation. 

After changes were made I ran tests for fastcore and fastai2 and all is working.


sgugger:
Don't forget to run `nbdev_install_git_hooks` before any PR. This one is hard to read because of all the metadata changes (and will cause many conflicts). Run `nbdev_clean_nbs` to force a clean up and push again, if you can.

rbracco:
Oops sorry about that. I ran both and ran tests again and pushed. Let me know if there's anything else needed. Thank you.

jph00:
You might need to run `nbdev_build_lib` as well, or just execute the last cell of the notebook.

rbracco:
Sorry not sure how I forgot that step. Should be good now.

jph00:
Many thanks!

PR #10, opened 2020-02-18
GetAttr cleanup and doc

pete88b:
- added docs and tests to help people understand `GetAttr#_xtra` https://forums.fast.ai/t/fastai-v2-chat/53518/108
- added \_component_attr_filter so we use the same logic in \_dir and \_\_getattr__
    - now \_dir only returns attributes that will be accessible
    - i.e. if \_xtra = ['lower'], you can't call .upper(), so I don't think it makes sense to have upper returned by \_\_dir__
    - to do this, we need to reference \_xtra from \_dir - which means \_xtra can't call \_dir - or we'll recurse forever
- changed `self._dir() if self._xtra is None else self._dir()` to `self._dir()`
    - i'm guessing this was/should have been `self._dir() if self._xtra is None else self._xtra` but as \_dir uses \_xtra in it's filter, we don't need a special condition for this now

I've not changed this but L#_xtra seems to be unused - might it make sense to remove this?


jph00:
Many thanks!

PR #13, opened 2020-02-27
Add torch.equal to the equals function.

KaiLicht:
The equals function is not using `torch.equal` to test if two tensors are equal (it does use `numpy.array_equal` for ndarrays). That results in pretty long testing times for a little larger tensors (e.g. a few seconds for `torch.zeros((3,800,1600))`.

### DIFF
--- a/fastcore/imports.py
+++ b/fastcore/imports.py
@@ -16,7 +16,9 @@ from uuid import uuid4
 
 # External modules
 import numpy as np
+import torch
 from numpy import array,ndarray
+from torch import Tensor
 from pdb import set_trace
 
 #Optional modules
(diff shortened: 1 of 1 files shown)


sgugger:
I don't think this should hurt. Thanks for the PR! I just edited to keep the vertical alignment inside that function.

mikonapoli:
Notice: this introduced a huge new dependency on pytorch and kinda broke fastdot

(cfr. https://forums.fast.ai/t/fastcore-fastdot-fastcore-added-torch-dependency-for-fastdot/64295 )

sgugger:
Ah yes, now I remember why we don't have this. `fastcore` does not depend on PyTorch so use `torch.equal` instead of `test_eq` (which is meant for small tensors only). Will revert.

PR #16, opened 2020-04-02
feat: add log_args to save function arguments

borisdayma:
As discussed in [fastai2 PR 225](https://github.com/fastai/fastai2/pull/225#issuecomment-607800612)


sgugger:
Thanks!

PR #17, opened 2020-04-04
fix(log_args): allow use of args with decorator

borisdayma:
`log_args` could be used as a function but as a decorator we could not pass arguments.
This is now fixed.

Additional tests have been added to the notebook.


sgugger:
Thanks for working on this!

borisdayma:
No problem, I think this is going to be really cool!

PR #18, opened 2020-04-04
feat: improve log_args

borisdayma:
* make log_args compatible with funcs_kwargs
* handle inheritance
* added test for `@funcs_kwargs`, `@delegates` and inheritance
* fail "silently" and output error


borisdayma:
I'm now adding the log_args to most interesting classes.
There's a lot of repetition in the `but` argument that could create issues.

What do you think if I do the following:

* when applied on a class or on `__init__`, I save `cls._log_args_but = but`
* For a class such as `TfmDL` or `WeightedDL`, I can use:
  * Option 1: `@log_args(but=DataLoader._log_args_but)` and I can still add more args in the but as `','.join([DataLoader._log_args_but, 'p1,p2'])
  * Option 2: `@log_args(but_as=DataLoader)`, which I will merge with any other but args I would give.

borisdayma:
Another note, it fails on `21_vision.learner.ipynb` at this line:

```python
dls = pets.dataloaders(untar_data(URLs.PETS)/"images", item_tfms=RandomResizedCrop(300, min_scale=0.5), bs=64,
                        batch_tfms=[*aug_transforms(size=224)], verbose=True)
```

It looks like we are not supposed to pass `item_tfms` based on the signature.
As a matter of fact, it seems that `RandomResizedCrop` does not appear anywhere so I believe it was not called correctly.

I'll try to take advantage of log_args to write a debug message and handle exceptions.

sgugger:
I think the option 2, `but_as`, feels more natural (I imagine it's for patched cases). Make sure it can accept a list cause we never know.

Otherwise, this looks good to me, thanks a lot!

borisdayma:
Good idea, it's actually for all functions using delegates like all the ones creating Learner which have same just of arguments we want to ignore.

Good idea to let it accept lists!

PR #22, opened 2020-04-06
Add classmethod patch option

erikgaas:
I was trying to solve a problem with respect to this thread: https://forums.fast.ai/t/splitting-a-class-with-static-methods/61794/5

This solution wound up working for me.


sgugger:
Oh thanks! Can't wait to try it!

PR #23, opened 2020-04-07
feat(log_args): log args before executing function

borisdayma:
### Issue

To keep the code minimal and avoid potential errors raised by `log_args`, we would first run decorated function and then log args.

The issue is that when we run `Learner.fit`, the function is ran prior to logging `fit` args so we would be missing those ones (since config is logged while executing this function).

### Solution

When `to_return=False` (default), we log the args prior to running decorated function.

When `to_return=True`, we have to log the args after running the function.

This solves the issue because it seems to be mainly related to `fit` functions and related, which all use `to_return=False`.

*Note*: this is a non-breaking change. It only let us log `fit` related args through callbacks.

### DIFF
--- a/fastcore/utils.py
+++ b/fastcore/utils.py
@@ -401,7 +401,6 @@ def log_args(f=None, *, to_return=False, but=None, but_as=None):
 
     @wraps(f)  # maintain original signature
     def _f(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


sgugger:
Sounds all reasonable to me. Thanks!

PR #24, opened 2020-04-08
fix(FixSigMeta): remove self from signature

borisdayma:
Remove `self` from signature using `FixSigMeta`.

This issue came from a user reporting a `log_args` error on `LabelSmoothingEntropy` (meaning that parameters on this function were not saved).

I noticed that the signature included `self`, then looked at `Module` which uses `PrePostInitMeta`, itself inheriting from `FixSigMeta`. Was tricky to solve but I think this should do it!

In theory there should never be a message from `log_args` unless there's a signature issue.

The only signature issues not shown are when too many parameters are passed because `**kwargs` is typically removed while actually everything is passed. In that case the parameter is named `function.parameter (not in signature)`.

### DIFF
--- a/fastcore/foundation.py
+++ b/fastcore/foundation.py
@@ -12,11 +12,17 @@ from .imports import *
 defaults = SimpleNamespace()
 
 # Cell
+def _rm_self(sig):
+    sigd = dict(sig.parameters)
+    sigd.pop('self')
+    return sig.replace(parameters=sigd.values())
+
+# Cell
 class FixSigMeta(type):
     "A metaclass that fixes the signature on classes that override __new__"
     def __new__(cls, name, bases, dict):
         res = super().__new__(cls, name, bases, dict)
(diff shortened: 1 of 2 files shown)


sgugger:
Did you check all the tests in fastcore pass with this change? I'm a bit nervous to change a function that is at the base of everything.

borisdayma:
Good point, I checked only the specific notebook but I should run all the tests of fastcore and fastai2.

My understanding of this specific line is that it's barely for displaying the help of the function. If we remove this line it shows something completely different since we overwrite the __new__ method.

I'll check all the tests and will get back to you.

sgugger:
It might impact other functions that rely on the signature, but if that notebook passes, all should be good. Just run make test in the repo to quickly test everything in parallel.

borisdayma:
Yes they all passed on fastcore!

sgugger:
Ok, let's merge this, thanks!

PR #26, opened 2020-04-10
fixes nested pipeline setup

lgvaz:
`Pipeline.setup` was not being called in a nested pipeline situtation:

```python
pipe1 = Pipeline([neg])
pipe2 = Pipeline([pipe1, neg])
```

pipe1 was being converted to a transform by `mk_transform`, overriding the `setup` method from the pipeline.


lgvaz:
I had to move `mk_transforms` tests down after the pipeline is defined. I don't like the current place, but I could not find a better location to put it.

lgvaz:
The current solution is still not functional. It breaks with the error: 
```python
TypeError: __call__() got an unexpected keyword argument 'split_idx'
```

Because `Pipeline.__call__` only receives one argument. We can certainly add `**kwargs` to it, but it might be dangerous.

lgvaz:
In case you're wondering why I want to nest two pipelines, it's because subclassed one of them and changed how it worked, because of this I cannot have all my transforms in a single pipeline.

lgvaz:
Okay, this is a **lot** more complicated than I originally thought... Pipeline delegates calls like `name` and `order` to it's functions, so I'm having to explicitly override these for this to work..


Maybe `Pipeline`s are not made for a nested structure =X

sgugger:
Most of the times, instead of having nested pipelines, I concatenate them (for instance in the datablock, if ` get_x` or `get_y`  are Pipelines, I extract the transforms from them) but I guess this won't work with your use case.

This sounds like a good first step in any case. We can deal with the `split_idx` in call and decode I guess, though I'd need to get a longer look at this. The order of the pipeline could default to 0 and its name could simply be its repr. Are there any other attributes that make this fail?

lgvaz:
I'll take a better look at this next week =)

PR #27, opened 2020-04-12
adds support for class dispatch

lgvaz:
Adds the ability to also dispatch on classes instead of only instances.

Very very useful for using type dispatched wrappers! (very cool)

### DIFF
--- a/fastcore/dispatch.py
+++ b/fastcore/dispatch.py
@@ -90,8 +90,10 @@ class TypeDispatch:
              for k in self.funcs.d for l,v in self.funcs[k].d.items()]
         return '\n'.join(r)
 
+    def _get_class(self, x): return x if inspect.isclass(x) else type(x)
+
     def __call__(self, *args, **kwargs):
-        ts = L(args).map(type)[:2]
+        ts = L(args).map(self._get_class)[:2]
         f = self[tuple(ts)]
         if not f: return args[0]
         if self.inst is not None: f = MethodType(f, self.inst)
--- a/nbs/03_dispatch.ipynb
+++ b/nbs/03_dispatch.ipynb
@@ -237,8 +237,10 @@
     "             for k in self.funcs.d for l,v in self.funcs[k].d.items()]\n",
     "        return '\\n'.join(r)\n",
     "\n",
+    "    def _get_class(self, x): return x if inspect.isclass(x) else type(x)\n",
+    "\n",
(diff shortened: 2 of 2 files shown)


sgugger:
Could you explain a bit more why you want to do this? I don't fully understand.

lgvaz:
Sure, what I'm trying to do is to inject functions that gets executed every time a `Transform` is called (kinda like hooks?), for that I did the following:
```python
@typedispatch
def subscribe(tfm:Transform):
    old_call = tfm.__call__
    def _inner(f):
        def _call(self, x, **kwargs):
            res = old_call(self, x, **kwargs)
            if res is not x: f(res)
            return res
        tfm.__call__ = _call
        return f
    return _inner
```
With this change I can subscribe to any `Transform` **class** and inject arbitrary functionality in the middle.
In my case, I'm developing a [library](https://github.com/lgvaz/pigboat) for weak supervision, so I want to inject these little functions that executes after transform and generate a task label, something like this:
```python
@labeller(Neg)
def func1(x): return 'cat2'

x = tensor(2)
tfms = Pipeline([Neg])
res = tfms(x)
```
```python
res.labels
>>> ['cat2']
```

Now, I may want `subscribe` to behave a little differently if I pass a `Transform` or another call to it.

---
For completeness I'll add the other necessary functions here:
```python
def _maintain_labels(old, new):
    if hasattr(old, 'labels'): new.labels = old.labels

_old_call = Transform._call
@patch
def _call(self:Transform, fn, x, **kwargs):
    res = _old_call(self, fn, x, **kwargs)
    _maintain_labels(x, res)
    return res

def _add_labels(f):
    def _inner(x):
        x.labels = getattr(x,'labels',[])
        x.labels.append(f(x))
    return _inner

def labeller(tfm):
    def _inner(f):
        return subscribe(tfm)(_add_labels(f))
    return _inner
```

sgugger:
> Now, I may want subscribe to behave a little differently if I pass a Transform or another class to it.

So you want a type dispatch that works on classes and not on objects. I don't think it should be the same function that does that (it seems a bit magical since I would expect the normal typedispatch to use its `type` implementation if I pass it `Transform`).
So I'd write a new `classtypedispatch` decorator (we can refactor the common behind the scenes in a private function) instead of having `typedispatch` be too magical.

lgvaz:
I will propose a PR with that =)

Issue #30, opened 2020-04-23
Should the test in __setitem__ be applied to __getitem__ and __delitem__ as well?


philtrade:
https://github.com/fastai/fastcore/blob/8a56bcbb2be3d5a2ab137b312c981a31d50f5cbe/fastcore/foundation.py#L281-L283

should `list(k) if isinstance(k,CollBase) else k` be applied to `__getitem__` (and potentially `__delitem__`) as well?

I was poking around [fastai2's `_42_tabular_rapids.ipynb`](https://github.com/fastai/fastai2/blob/master/nbs/_42_tabular_rapids.ipynb).    In [`Tabular.reduce_cats()`](https://github.com/fastai/fastai2/blob/48ba39ac92d5fa4e4d86013797e100d6c3be2aea/fastai2/tabular/core.py#L139), the index `k` is an `L`-list, which `cudf.DataFrame` can't handle.   The following would fail and point to the above line, where `self.items` is a `cudf.DataFrame`.  Listifying `k` would get pass this hiccup:
```
df = cudf.from_pandas(pd.DataFrame({'a':[0,1,2,0,2]}))
to = TabularGPU(df, Categorify, 'a')
```

sgugger:
This sounds reasonable. Do you mind making a PR with it?

sgugger:
Let's say it's fixed by #32, if you want to add del in a separate PR I'll review it but you seemed to say it was more complicated.

philtrade:
Thank you for merging.  Yeah I don't wanna jump the gun on `del(x[ [slice] ])`.  Perhaps later, when I'm more comfortable w/ the usage cases by the derived classes of `CollBase`.

PR #31, opened 2020-04-23
CollBase.__getitem__(): listify index like __setitem__() does

philtrade:
Per issue #30.


philtrade:
will retry from a clean slate....  might have the wrong `nbdev` in my dev environment (again)....

PR #32, opened 2020-04-23
CollBase.__getitem__(): listify index like __setitem__().

philtrade:
(resubmitting, after updating `nbdev`)

Per issue #30.

Only fix `__getitem__()` for now, coz' `__delitem__()` seems more subtle (multiple rows indices? column names?) -- rather keep it simple at the moment.


sgugger:
Thanks!

PR #34, opened 2020-04-28
Add classmethod delegation

erikgaas:
I ran into a problem where I wanted to be able to delegate the kwargs of one class method within a class to a different classmethod within the same class.

This was an issue because inspect.signature would not work on a classmethod because it is not callable. All I needed to do was add a line that made the inspection of to_f to be the same as from_f.

Looks like some extra uncommited documentation additions creeped into this PR as well.


sgugger:
Thanks, this looks useful!

PR #41, opened 2020-07-14
updated store_attr so nms can be None

pete88b:
Hopefully this will help save a little typing allowing things like
```python
class Interpretation():
    "Interpretation base class, can be inherited for task specific Interpretation classes"
    def __init__(self, dl, inputs, preds, targs, decoded, losses):
        store_attr(self, "dl,inputs,preds,targs,decoded,losses")
```
to be written as
```python
class Interpretation():
    "Interpretation base class, can be inherited for task specific Interpretation classes"
    def __init__(self, dl, inputs, preds, targs, decoded, losses):
        store_attr(self)
```
Hope it also helps that store_attr can now pick up parameters from kwargs too.

### DIFF
--- a/fastcore/utils.py
+++ b/fastcore/utils.py
@@ -80,10 +80,13 @@ class ignore_exceptions:
     def __exit__(self, *args): return True
 
 # Cell
-def store_attr(self, nms):
+def store_attr(self, nms=None):
     "Store params named in comma-separated `nms` from calling context into attrs in `self`"
     mod = inspect.currentframe().f_back.f_locals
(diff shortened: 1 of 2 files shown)


jph00:
Sorry didn't notice this - I've actually already added a way to not require params, but it relies on a particular field name. Feel free to re-submit a PR similar to yours which takes account of this change if you want.

jph00:
BTW please either at-mention me or request me as a reviewer on your PRs, to ensure I don't miss them.

PR #44, opened 2020-08-14
Moves nested functions out of parallel_gen

HenryDashwood:
Hello! 

This pull request is based on the solution [offered here Gabriel Torrealba](https://forums.fast.ai/t/bigdata-databunch-and-training-loop/60003/32) to `parallel_gen` not being pickle-able.

It solved the problem for me and all the tests still pass so I think it's an acceptable thing to do. It's quite possible that I've missed a good reason for the existing structure of the code though in which case, my apologies! 

I think I've cleaned the notebooks appropriately but do let me know if there is still some undesirable stuff in there.


HenryDashwood:
Oh that's not good

jph00:
Many thanks. We had a conflict and also we need the funcs to be private, so I've made that change directly on master.

PR #46, opened 2020-08-19
more docs on 02_utils

hamelsmu:
I rearranged some of the functions into different sections, so this is why the diff looks so crazy


jph00:
@hamelsmu looks good - just a build failure for you to look at.

PR #47, opened 2020-08-19
Update Makefile to allow for local preview

hamelsmu:
it only works when baseurl "", which we can override thankfully in the CLI


jph00:
What's not working for you without this?

jph00:
I notice the sidebar doesn't work. Anything else?

jph00:
I've just pushed a change which should make this unnecessary. Lemme know if it works for you.

PR #48, opened 2020-08-20
Clean up docs for utils

hamelsmu:


hamelsmu:

hamelsmu:
@jph00 I made the following changes:

- removed `show_doc` from all functions, which got rid of the vast majority of `show_docs`
- I kept `show_doc` on most classes to indent them to level 4 to be peers of functions within the same group
- I kept `show_doc` in special cases where I needed to hide something, or where I needed show_doc specifically to accomplish something.  

LMK if that works

hamelsmu:
Actually give me a moment, I'll tag you again I just realized I can get rid of more `show_doc`

hamelsmu:
@jph00 ok this is now ready for review.

jph00:
Great!

PR #51, opened 2020-08-23
add prose to FixSigMeta

hamelsmu:
Made the following changes:

- Moved the test utility `test_sig` to the tests module.
- Fleshed out the prose for `FixSigMeta`

I would normally do a larger PR with more changes, however I just want to make sure my style conforms to fastai style so I decided to have you review this particular change in order to make sure I have the right idea.  

I know you said to only mention the situation where `__call__` overrides the signature in passing, however, I found that this scenario is utilized in this library and was key to understanding what is happening, so I left it in to get your opinion.  Happy to modify this if you like.


hamelsmu:
Ready for review @jph00

jph00:
Nice!

PR #52, opened 2020-08-24
Add Integration tests for fastai into CI

hamelsmu:
These tests also check to make sure that any changes to fastcore do not break anything in fastai

@jph00 ready for review - the length of CI, however, increases from 2 to 4 minutes

We could alternatively make these integration tests nightly


jph00:
This is great :)

PR #55, opened 2020-08-27
Add more docs to fastcore.foundations

hamelsmu:
These are just the ones I lost today :) Wohoo I rewrote it though


hamelsmu:
@jph00 ready for review

jph00:
This is great! And so fast 😎

PR #58, opened 2020-08-28
fallback behavior for @typedispatch

antorsae:


antorsae:
Well this does not work, strange if I run (w/ this PR):

```
@typedispatch
def to_detach(learn:(Learner,NoneType),b:object,cpu:bool=True,gather:bool=True):
    return learn.dl.to_detach(b,cpu,gather) if hasattr(getattr(learn,'dl',None),'to_detach') else _to_detach(b,cpu,gather)
```
and then:
```
to_detach((tensor([1.]).cuda(),tensor([2.])))
```
The regular fallback (defined in 00_torch_core) `to_detach` is NOT executed and I just get the first arg back:

```
(tensor([1.], device='cuda:0'), tensor([2.]))
```

The reason seems to be `to_detach` is not in `globals()` when @typedispatch is executed.

No idea what's going on.

antorsae:
Closing as this works within the module where Typedispatch is instantiated but does not work elsewhere (`globals()` behaves differently and I have no idea why).

PR #61, opened 2020-08-29
Fix _config.yml b/c its causing links to break

hamelsmu:
This fixes broken links to colab notebooks ready for review @jph00


jph00:
Oops!

PR #62, opened 2020-08-29
Finish foundation (first pass)

hamelsmu:
I decided to hide the following two items which are just there in support of everything else and don't seem useful to document:

- function `custom_dir`
- class `CollBase`

This is my first pass and fleshing out the docs for foundation, and will move onto a different section.  I added some docs to `L` and I think that section is fairly documented.   While I could spend a great more deal of time on `L`, I don't think it is worth getting too bogged down in it as I think its fairly approachable from the tests.  I might come back to L on a second pass through the docs if I have time.  But do let me know what you think, especially if you think particular method(s) needs some special love. 

Ready for review @jph00

### DIFF
--- a/nbs/01_foundation.ipynb
+++ b/nbs/01_foundation.ipynb
@@ -2735,7 +2735,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "`ColBase` is a base class that emulates the functionality of python `list`:"
+    "`ColBase` is a base class that emulates the functionality of a python `list`:"
    ]
   },
   {
--- a/nbs/01_foundation.ipynb
+++ b/nbs/01_foundation.ipynb
@@ -1418,6 +1418,26 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `method` function is used by `func_kwargs` to optionally change a function's type to a method.  In the below example we change the type of `a` from a function to a method:"
+   ]
+  },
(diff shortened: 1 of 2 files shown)


nbs/01_foundation.ipynb
```diff
@@ -1417,6 +1417,26 @@
     "    return MethodType(f, 1)"
```
jph00, line 1417: <p>Remove "by func_kwargs" since if it were only to support that, it would be private (i.e prefixed by <code>_</code>).  It's odd that python doesn't have builtin simple way to mark something as a method, so this is a really useful and important workaround for that in general.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/fastcore/pull/62/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='109'/>

jph00:
Best to unhide those 2 @hamelsmu . They're not just to support `L` (if they were, they'd be private) - they're really rather useful tools in the toolbox, IMO! :)

hamelsmu:
> Best to unhide those 2 @hamelsmu . They're not just to support L (if they were, they'd be private) - they're really rather useful tools in the toolbox, IMO! :)

Ok!  I figured that means that I should write some docs and tests for them, too!  Which I did!  Thanks @jph00  its ready for review now!

jph00:
Splendid!

PR #69, opened 2020-09-05
fix store_attr would use 'names' as 'self'

shrimpceviche:
When using the old version, when not properly passed, the ```name``` would be ```self```. The update will fix it.


jph00:
No this isn't how we want the function to work.

PR #77, opened 2020-09-12
feat(store_attr): handle functions vs methods

borisdayma:
I tried to find a more robust way to know if I was in a function vs method by trying to access full qualified name (where we could just look for the dot).
However it does not seem to be accessible this way: https://bugs.python.org/issue13672

Here an easy workaround would be to assume that any class method always call the first argument as "self", which should work at least for the `fastai` library (and probably in most other libraries).

### DIFF
--- a/fastcore/utils.py
+++ b/fastcore/utils.py
@@ -98,7 +98,7 @@ def store_attr(names=None, self=None, but=None, **attrs):
     if not hasattr(self, '__stored_args__'): self.__stored_args__ = {}
     if attrs: return _store_attr(self, **attrs)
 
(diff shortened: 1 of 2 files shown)


borisdayma:
Looks clean with nbdev 1.0.18.
Should I install from git?

borisdayma:
@jph00 Let me know what you think of this approach.
If it's ok, then I think there are all the elements in place to start replacing `log_args` in fastai

jph00:
Thanks! I just implemented this, but a slightly different way. Dunno if it's more or less hacky... ;)

Issue #100, opened 2020-09-27
Inconsistency with cmp_instance and typedispatch because of binary sort algorithm


kessido:
Basically, typedispatch assume that the type in `funcs` are ordered in a topological sort, but this is not true as the key comparison is not defined well under this.

stupid example (edited, now it is less stupid :P ):

```
>>> from fastcore.dispatch import *
>>> 
>>> class A: pass
... 
>>> class B(A): pass
... 
>>> @typedispatch
... def f(a:A): print('A')
... 
>>> @typedispatch
... def f(x:int): pass
... 
>>> @typedispatch
... def f(x:tuple): pass
... 
>>> @typedispatch
... def f(b:B): print('B')
... 
>>> @typedispatch
... def f(x:list): pass
... 
>>> f(B()) 
A
>>> print(f)
(list,object) -> f
(tuple,object) -> f
(A,object) -> f
(int,object) -> f
(B,object) -> f
```

doing insertion sort variation with the key comparison at hand will solve the issue I believe. I can implement this quickly but would love hearing your opinion first.

Ido

jph00:
I'd love to see your approach as a PR - honestly, I didn't put that much time into thinking about the sort, and just kinda used the first thing I tried because it passed my tests! I'm glad you've taken the time to think through it more carefully.  A test showing where the existing approach breaks would be very useful to include in the PR.

PR #101, opened 2020-09-28
Fix _TypeDict types topological ordering

kessido:
This pull request fixes 2 things.
1. Issue #100 where dispatch might miss possible subclass because of wrong ordering.
2. Not known issue with typing.Collection\typing.List\etc. where using them as first argument to is issubclass might throw exception (added example in notebook).


jph00:
Great stuff! :)

Issue #103, opened 2020-10-01
Can't use classmethod decorator!


Salehbigdeli:
Some users may expect to use `typedispatch` decorator on top of `classmethod` or `staticmethod` they can't right now. I'm going to fix this.
Failing example:
```
class A:
    @typedispatch
    def f_td_test(self, x:numbers.Integral, y): return x+1
    @typedispatch
    @classmethod
    def f_td_test(cls, x:int, y:float): return x+y
    @typedispatch
    @staticmethod
    def f_td_test(x:int, y:int): return x*y
    
test_eq(A.f_td_test(3,2), 6)
test_eq(A.f_td_test(3,2.0), 5)
test_eq(A().f_td_test(3,'2.0'), 4)
```

PR #106, opened 2020-10-02
Make typedispatch decorator more general

Salehbigdeli:
Making `typedispatch` more general in order to use with `staticmethod` or `classmethod` decorator. More detail in issue #103


jph00:
Very nice :)

Issue #111, opened 2020-10-07
Minor bug in _oper


Salehbigdeli:
In the definition of `_oper` there is an assumption about the third argument. If it's not equal to the default it assumes that we want to use `op` in partial case.

Failing example:
```
assert math.isnan(add(1, float('nan')))
```
We also can't use `None` as default in `_oper`, at least for getting more useful error messages.

PR #112, opened 2020-10-07
Fix default in oper

Salehbigdeli:
The solution to issue #111


jph00:
Thanks! Sorry I just re-organized this file and created a conflict, so I'll add the fix manually now.

jph00:
(FWIW `fastrelease` will still ensure that you get credit in the release notes :) )

PR #127, opened 2020-10-12
change name to ExceptionExpected

hamelsmu:


hamelsmu:
@jph00 ready for review changed he name :)

PR #134, opened 2020-10-17
Fix copy_func to include required keyword defaults

worc3131:
```
def g(x, *, y=3):
  return x+y
copy_func(g)(4)
```
This code currently throws
```
g() missing 1 required keyword-only argument: 'y'
```
This is because the keyword defaults are being lost. It doesn't seem that these can be passed in `__init__` of the function class, so this PR populates them after. This is only an issue on required keyword arguments with defaults as non required keyword arguments are stored in `__defaults__`.


jph00:
Many thanks

Issue #143, opened 2020-10-22
L(x, use_list=None) doesn't make a list if x is a numpy.float64


hal-314:
L(x, use_list=None) behaves differently if you pass a float or a numpy.float64/numpy.float32, etc

To reproduce:
```
print(L(np.asarray([7.,5.], dtype=np.float64)[0]))
print(L(np.asarray([7.,5.], dtype=np.float32)[0], use_list=None))
print(L(7., use_list=None))
```
Prints:
```
[7.0]
7.0
[7.0]
```

I found this issue then using Optuna 2.2 to optimize a model (it didn't happen with Optuna 2.1). Optimizer couldn't set the value. Here is the error stack:
```
> /miniconda3/envs/pytorch/lib/python3.7/site-packages/fastcore/foundation.py(271)__len__()
    269     "Base class for composing a list of `items`"
    270     def __init__(self, items): self.items = items
--> 271     def __len__(self): return len(self.items)
    272     def __getitem__(self, k): return self.items[list(k) if isinstance(k,CollBase) else k]
    273     def __setitem__(self, k, v): self.items[list(k) if isinstance(k,CollBase) else k] = v

ipdb>  self.items

0.00038467256590927996

ipdb>  type(self.items)

<class 'numpy.float64'>

ipdb>  u

> /miniconda3/envs/pytorch/lib/python3.7/site-packages/fastai/optimizer.py(43)set_hyper()
     41             else: v = [v.stop/10]*(len(self.param_lists)-1) + [v.stop]
     42         v = L(v, use_list=None) # <-- Here, v isn't a list or something with len() method
---> 43         if len(v)==1: v = v*len(self.param_lists)
     44         assert len(v) == len(self.hypers), f"Trying to set {len(v)} values for {k} but there are {len(self.param_lists)} parameter groups."
     45         self._set_hyper(k, v)


```

Salehbigdeli:
The problem comes from `_is_array` function, this function checks if given argument is `numpy.ndarray` , `pandas.DataFrame` or `pandas.Series`. But unfortunately it simply checks with attributes not exact `shape` of the given parameter.
I'm going to fix it.

The problem:
```
#export
def _is_array(x): return hasattr(x,'__array__') or hasattr(x,'iloc')
```

My solution:
```
#export
def _is_array(x): 
    if hasattr(x,'__array__') or hasattr(x,'iloc'): return bool(x.shape)
```

hal-314:
[...]

PR #146, opened 2020-10-23
Fix for issue #143

Salehbigdeli:
See my comment on issue #143

### DIFF
--- a/fastcore/foundation.py
+++ b/fastcore/foundation.py
@@ -275,6 +275,10 @@ class CollBase:
     def __delitem__(self, i): del(self.items[i])
     def __repr__(self): return self.items.__repr__()
     def __iter__(self): return self.items.__iter__()
+    @property
+    def shape(self):
+        if hasattr(self.items, 'shape'): return self.items.shape
+        return np.array(self.items).shape
 
 # Cell
 class _L_Meta(type):
--- a/fastcore/foundation.py
+++ b/fastcore/foundation.py
@@ -114,7 +114,8 @@ def mask2idxs(mask):
     return [int(i) for i in mask]
 
 # Cell
-def _is_array(x): return hasattr(x,'__array__') or hasattr(x,'iloc')
+def _is_array(x):
+    if hasattr(x,'__array__') or hasattr(x,'iloc'): return bool(x.shape)
 
 def _listify(o):
     if o is None: return []
--- a/fastcore/foundation.py
+++ b/fastcore/foundation.py
@@ -115,7 +115,8 @@ def mask2idxs(mask):
 
 # Cell
 def _is_array(x):
-    if hasattr(x,'__array__') or hasattr(x,'iloc'): return bool(x.shape)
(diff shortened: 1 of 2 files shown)


Salehbigdeli:
More detail on these changes can be found in issue #147

Salehbigdeli:
I don't know if I broke something in plots! Since we don't have anything to check plots automatically, here is screenshot of cell that made last pipeline execution fail:

image

fastcore/foundation.py
```diff
@@ -114,7 +114,8 @@ def mask2idxs(mask):
     return [int(i) for i in mask]
 
 # Cell
-def _is_array(x): return hasattr(x,'__array__') or hasattr(x,'iloc')
+def _is_array(x):
+    if hasattr(x,'__array__') or hasattr(x,'iloc'): return bool(x.shape)
```
hal-314, line 118: I think that you broke things as you return `None` is it doesn't enter in the `if`. So, you could rewrite as:
```
def _is_array(x):
    # First, check if it could be an array (np.array or pandas Dataframe/Serie and then check that has shape. If it doesn't , it's a constant value. 
    return (hasattr(x,'__array__') or hasattr(x,'iloc')) and bool(x.shape)
```

hal-314 (changes requested):

fastcore/foundation.py
```diff
@@ -274,6 +276,11 @@ def __setitem__(self, k, v): self.items[list(k) if isinstance(k,CollBase) else k
     def __delitem__(self, i): del(self.items[i])
     def __repr__(self): return self.items.__repr__()
     def __iter__(self): return self.items.__iter__()
+    @property
+    def shape(self):
+        if hasattr(self.items, 'shape'): return self.items.shape
+        import numpy as np
```
jph00, line 282: We can't introduce a dep on numpy into fastcore. fastcore is designed to be lightweight with no deps outside the standard lib.

fastcore/foundation.py
```diff
@@ -114,7 +114,9 @@ def mask2idxs(mask):
     return [int(i) for i in mask]
 
 # Cell
-def _is_array(x): return hasattr(x,'__array__') or hasattr(x,'iloc')
+def _is_array(x):
+    "First, check if it could be an array (np.array or pandas Dataframe/Serie and then check that has shape. If it doesn't , it's a constant value."
```
jph00, line 118: The "First, " here is confusing. I'd suggest removing it

jph00 (changes requested):
I think the piece where you are using numpy needs more thought, and maybe some tests/examples. When exactly would you end up in that part of the conditional? Would it always be true, or always false? What do you need to check for, really?

fastcore/foundation.py
Salehbigdeli: Ok, I'll remove it.

jph00:
Seems like this is obsolete now. Please close obsolete PRs in the future @Salehbigdeli, since it helps keep things clean! :) If for any reason you do need to create a new PR instead of updating an existing one, be sure to reference the old closed PR from the new one.

PR #156, opened 2020-10-28
Add doc to Transform meta

tcapelle:
This pull request add documentation to the `Transform` class, exaplining the usage of the `TfmMeta` metaclass, it is based on [Jeremy's walk thru 05](https://www.youtube.com/watch?v=GcMGchBJeXk&feature=youtu.be) and for my own understanding.


tcapelle:
why something that did not change any code don't pass the CI?

jph00:
Many apologies for the CI fail. It's due to the new PyTorch - you can ignore it.

Thanks for these nice docs. I'm not quite sure what to do with them, because they're documenting a private symbol. So `_TfmMeta` doesn't appear in the docs at all, and it would be quite confusing for docs readers to find that they're reading about a class that they never even saw! Any thoughts?...

tcapelle:
You are entirely right!
I mostly did this to understand what is happening behind the
`Transform/Pipeline`.

jph00:
On reflection, I'm going to close this, because it should indeed be private, so shouldn't be in the docs. Feel free to do a PR with some basic comments in the code if you think it would help.

PR #167, opened 2020-11-05
Test_eq does not work for pandas + Test_eq_type is not correct for inner types

kessido:
i.e `test_eq_type([[1]], [[1.]])` will not raise any exception, as it only check the outside type.

Added tests + support for other types.


hamelsmu:
Can you add some more prose in the notebooks to illustrate what these changes add?  I'm not sure I completely follow

kessido:
I think I will split it into two for it to be more understandable.

PR #169, opened 2020-11-06
add __repr__ to Param for augmented function documentation

rsomani95:
I'd like to propose adding a `__repr__` to Param, as I think it allows for nicer documentation of functions being used interactively in a notebook.

```python
from fastcore.script import Param as P

def func(a: P("first param", int) = 10,
         b: P("second param", int) = 20):
    print(a+b)
```

When using this function inside a notebook, and looking up the docs with `<Shift + Tab>`, I get to see the `help` included in the function signature.

(EDIT): For completeness, this is what the function signature look like:
```python
func(a: int <first param> = 10, b: int <second param> = 20)
```


fastcore/script.py
```diff
@@ -43,6 +51,11 @@ def pre(self): return '--' if self.opt else ''
     @property
     def kwargs(self): return {k:v for k,v in self.__dict__.items()
                               if v is not None and k!='opt' and k[0]!='_'}
+    def __repr__(self):
```
rsomani95, line 54: The help is not returned as a string, but just printed out enclosed in `<...>`. As a result, syntax highlighting is applied. Perhaps there's a cleaner way to refactor this `if-else` party too

I'll readily change this per your suggestions

jph00:
Many thanks! I've been experimenting with something similar recently as it happens :) I might end up replacing your implementation with the results of my experiments later, but very happy to use your PR, at least for now.

rsomani95:
Oh cool! Glad to hear that

I've been going back and forth between placing the doc help inside single quotes as syntax highlighting is applied otherwise. Also, `choices`, if any, are not currently displayed. I'm sure you've thought of all this; I'm very curious to see your take on it :)

PR #184, opened 2020-11-10
make sure typed works with classes

### LINKED ISSUE #183: @typed doesn't work with classes: key error on self
getting the error that `self` was not found in dict named `anno`

This bug was reported by @krishnap

hamelsmu:
fixes #183


hamelsmu:

jph00:
Thanks!

PR #188, opened 2020-11-11
Fix pandas test eq

kessido:
Hey, I notice that pandas test_eq is incorrect.
I added a test to see that it always return True, and that it doesn't actually fail when given wrong inputs.
Instead I added a call for .equals which is implemented in pd.Series\pd.DataFrame.

This resulted in a failing test now, which I fixed.

Ido


kessido:
integration test fail because it should P: pandas doesnt support this comparison.

```
>>> import pandas as pd
>>> a = pd.Series([1,2,3,4])
>>> b = [1,2,3,4]
>>> a.equals(b)
False
```

PR #189, opened 2020-11-11
Fix test eq type for inner type

kessido:
test_eq_type did not work for inner types. simple example:
`test_eq_type([[1]],[[1.0]])` does not fail.

I implemented a recursive comparison instead.


jph00:
That's a good idea, but I think it's trying to do too much. As a result, performance is bad for tensors/arrays/etc, since listifying them is really slow. Also, this implementation is a bit more complicated than I'm comfortable with.

Arrays etc should already be testing the inner type, shouldn't they? So maybe that stuff isn't needed?

jph00 (changes requested):
Simplify and improve performance.

kessido:
> Simplify and improve performance.

Working on it now.

kessido:
Simplified&Improved performance. @jph00 please review :)

> That's a good idea, but I think it's trying to do too much. As a result, performance is bad for tensors/arrays/etc, since listifying them is really slow. Also, this implementation is a bit more complicated than I'm comfortable with.
> 
> Arrays etc should already be testing the inner type, shouldn't they? So maybe that stuff isn't needed?

Arrays doesn't test inner types, so it's still needed.
Also for cases where np.array uses object, only testing dtype can be misleading.

hamelsmu:
@kessido can you resolve conflicts?

jph00:
Closing this since it's gotten stale, and I'm still not sure this is actually behavior we want.

PR #190, opened 2020-11-11
fixed equal for tensors

kessido:
Equals does not work well for tensors, and fails when presented with tensor with different sizes.
Added tensor to use `array_equals` instead, and added tests.

Ido


jph00:
Thanks!

PR #191, opened 2020-11-11
Proposition: optional override to is_instance(x,t) to support both types and str in the same function

kessido:
splitting the is_instance, and is_instance_str seemed like a complicated solution imho, I'm not too keen with the override, but I find it much more appealing in the later uses imo.

Then again, just a proposition.

### DIFF
--- a/fastcore/dispatch.py
+++ b/fastcore/dispatch.py
@@ -174,7 +174,7 @@ def default_set_meta(self, x, as_copy=False):
 def cast(x, typ):
     "cast `x` to type `typ` (may also change `x` inplace)"
     res = typ._before_cast(x) if hasattr(typ, '_before_cast') else x
-    if isinstance_str(res, 'ndarray'): res = res.view(typ)
+    if isinstance(res, 'ndarray'): res = res.view(typ)
     elif hasattr(res, 'as_subclass'): res = res.as_subclass(typ)
     else:
         try: res.__class__ = typ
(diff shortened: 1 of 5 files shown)


jph00:
I think this is a good idea, but let's change the definition of `rinstance`, which we already have, instead of replacing `isinstance`. Does that sound OK?

kessido:
> I think this is a good idea, but let's change the definition of `rinstance`, which we already have, instead of replacing `isinstance`. Does that sound OK?

Sounds ok
I find it less ram consuming (metaphorically) to add it to the base function as another functionality~

kessido:
> I think this is a good idea, but let's change the definition of `rinstance`, which we already have, instead of replacing `isinstance`. Does that sound OK?

Another option btw, is to combine both risinstance and isinstance. e.g:
https://github.com/kessido/kesscore/blob/3657275b4e69cf708a5bcdbde6ab7afd583cdda9/kesscore/functional.py#L45
but that just a matter of taste, I find it more appealing to write:
`L(1,2,3).map(isinstance(int)).all()`

hamelsmu:
@kessido where did you end up on this PR?  Just want to follow up.   I see some discussions about changing `rinstance` but not sure I see any commits after this.  Thanks

jph00:
Since this had gotten rather stale I implemented it myself. `risinstance` now handles strings and lists of strings. `isinstance_str` is still there, but just to support `array_equal`.

PR #220, opened 2020-11-24
Cast

tcapelle:
- Add `casted` decorator to enforce casted output type
- Moved `with_cast` to `04_dispatch` with the other casting functions.


jph00 (changes requested):
Thanks @tcapelle . Don't move `with_cast` - that's a breaking change, and also `dispatch` has more imports so is slower to use. Also, don't add a new `casted` function, instead change `with_cast` to cast the return value if a return annotation is present.

jph00:
(Looks like there's some conflicts to resolve BTW.)

tcapelle:
Hey, sorry for moving the `with_cast`, seemed logical.
In my use case, the function that I am `casting` is a `TypeDispatch` one, so I don't want to `cast` params, only output. Mostly because I am using opencv with `ArrayImage` and getting back `np.array`, so want to cast back the output to image.

PR #228, opened 2020-11-25
add cast to output

tcapelle:
In `with_cast`:
- Add cast to output if annotation is present

Question to @jeremy : With `TensorImage` and `ArrayImage` one has to call `cast` instead of directly calling the class name (as the `__init__` is not implemented), should not we use cast in `with_cast` instead of calling the constructor directly?


hamelsmu:

jph00:
Great job!

PR #257, opened 2020-12-16
Fixes #256

### LINKED ISSUE #256: TypeError with progress=True in fastcore.parallel.parallel
I'd installed fastcore 1.3.12 but hadn't installed fastprogress and then ran the following
```
def return_one(i):
    return 1
parallel(f=return_one, items=range(5), n_workers=2, progress=True)
```

This leads to a stack trace
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-25-4910dc0c904f> in <module>
      3 def return_one(i):
      4     return 1
----> 5 parallel(f=return_one, items=range(5), n_workers=2, progress=True)

~/anaconda3/envs/python3/lib/python3.6/site-packages/fastcore/parallel.py in parallel(f, items, n_workers, total, progress, pause, threadpool, timeout, chunksize, *args, **kwargs)
    104         if progress:
    105             if total is None: total = len(items)
--> 106             r = progress_bar(r, total=total, leave=False)
    107         return L(r)
    108 

TypeError: 'NoneType' object is not callable
```

I believe there is an error in [this line](https://github.com/fastai/fastcore/blob/master/fastcore/parallel.py#L100) in that it calls us to reach 106 even if `progress_bar` is `None`

I'll submit a patch for this as well

dansbecker:
I believe the desired behavior if a user doesn't have fast_progress installed is to continue with no progress bar rather than to throw an exception... though a warning would also be reasonable.

Here's a patch that continues without the progress bar in that case.

### DIFF
--- a/fastcore/parallel.py
+++ b/fastcore/parallel.py
@@ -97,11 +97,10 @@ except: progress_bar = None
 def parallel(f, items, *args, n_workers=defaults.cpus, total=None, progress=None, pause=0,
              threadpool=False, timeout=None, chunksize=1, **kwargs):
     "Applies `func` in parallel to `items`, using `n_workers`"
-    if progress is None: progress = progress_bar is not None
     pool = ThreadPoolExecutor if threadpool else ProcessPoolExecutor
     with pool(n_workers, pause=pause) as ex:
         r = ex.map(f,items, *args, timeout=timeout, chunksize=chunksize, **kwargs)
-        if progress:
+        if progress and progress_bar:
             if total is None: total = len(items)
             r = progress_bar(r, total=total, leave=False)
         return L(r)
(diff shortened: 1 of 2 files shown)


hamelsmu:
Hi @dansbecker you have to actually edit the corresponding notebook in order to change this code.  I'm happy to pair with you if you are interested!  So nice to see you here! ❤️ 

cc: look@jph00 its our friend, @dansbecker !!!

dansbecker:
@hamelsmu Ah... I assume I am supposed to make the change [here](https://github.com/fastai/fastcore/blob/master/nbs/03a_parallel.ipynb) and not directly in the .py file?

I'm excited for any excuse to catch up, but I have a hunch this change will be a 1-minute change when we pair up.

jph00 (changes requested):
Thanks @dansbecker . If you just run `nbdev_build_lib` now it should be good to go!

dansbecker:
Thanks @jph00. 

Your comment caused me to look at nbdev a little more... it's cool to see what's happening here.

jph00:
Thanks @dansbecker - you might be interested in this video:
https://www.youtube.com/watch?v=9Q6sLbz37gk

PR #258, opened 2020-12-18
Some small ergonomic fixes for running with docker-compose

whoahbot:
Just a few things I found when first booting up fastcore in VS Code with the remote containers extension.

- Sets the workspace folder to /data instead of / inside the container.
- Install the dev dependencies (matplotlib) when booting the notebook container.


hamelsmu (approved):
thanks!

PR #259, opened 2020-12-18
neptune-cli -> neptune-client

whoahbot:
Not sure if neptune-cli is used anywhere, but I believe that the neptune-cli client was renamed to neptune-client.

https://github.com/neptune-ai/neptune-client

This should resolve the build errors I'm seeing:

```
ERROR: Requested neptune-cli from https://files.pythonhosted.org/packages/ba/ae/0cdcb9d20bf3b2324d21cfd11dc05246972e9759e7609739c6ce0ef01639/neptune-cli-2.8.23-1.tar.gz#sha256=343bd3d79da2c2fd10b9cd8ffb52570f8eb91573dbf290742eefb1aa643fe2eb has different version in metadata: '2.8.23'
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
         pip install -Uqq nbdev
         cd fastai && pip install -Ue .[dev]
         cd ../fastcore && pip install -Ue .[dev]
-        pip install "sentencepiece<0.1.90" wandb tensorboard albumentations pydicom opencv-python scikit-image pyarrow kornia catalyst captum neptune-cli
+        pip install "sentencepiece<0.1.90" wandb tensorboard albumentations pydicom opencv-python scikit-image pyarrow kornia catalyst captum neptune-client
 
     - name: check for cache hit
       uses: actions/cache@v2


hamelsmu:
Thanks this is great!

hamelsmu:
cc: @jph00

jph00:
Many thanks for noticing this @whoahbot . To explain, there are a number of optional 3rd party integrations in fastai - Neptune is one of them. It's optional, unless you want to use Neptune. But it's needed to build the docs, since the notebook has to be able to run.

There's actually a set of dev_requirements in fastai's settings.ini, which is really the right place for this fix, so I've made that fix now, and removed the extra pip install line from the workflow entirely.

I would prefer to move every 3rd party integration into a separate repo BTW. If you're interested, that could be a useful project to work on. Happy to discuss.

jph00:
Great! Thanks for making progress on this. The next step is to make it so this can install as `fastai.callback.neptune`. Info to get started on this is here:

https://packaging.python.org/guides/packaging-namespace-packages/

`_nbdev.py` isn't going to work, so nbdev probably needs some bool config setting like `extension` which, if true, means this is a namespace extension module so _nbdev.py, Makefile, _config.yml, etc shouldn't be built/modified.

(We need a better solution for this, which we'll look at in nbdev v2.)

PR #276, opened 2021-01-12
add ceil to make mx and mn work in sparkline

### LINKED ISSUE #277: Sparkline errors when a value is given that exceeds `mx`
image

hamelsmu:
Fixes bug in sparkline when a maximum value is seen greater than `mx`

closes #277


jph00:
lgtm, assuming that the test you added fails with the previous code (i.e it'll identify any regressions in the future). looks like you need to clean nbs btw

PR #283, opened 2021-01-21
support windows: only add lock if it runs parallelly

mszhanyi:
1. support windows: 00_test and 03_xtras are not compatible with Windows
2. Fix https://github.com/fastai/nbdev/issues/393
If the functions runs sequentially, it doesn't need a lock. 
And multiprocess.manager().lock() spawns one new process, notebook will be stuck on Windows.


mszhanyi:
@jph00 could you please take a look?

jph00:
Great! :)

jph00:
It's currently marked as a draft. Please mark it as ready to review if you want it merged.

mszhanyi:
> It's currently marked as a draft. Please mark it as ready to review if you want it merged.



> It's currently marked as a draft. Please mark it as ready to review if you want it merged.

@jph00 ， I marked it as draft because I need more time to enable all fastcore support windows.
Now. all tests are passed in windows.
image

mszhanyi:
FYI.
1. nbdev_clean_nbs always update \r to \n in notebooks. 
   so I have to remove as_bytes=True in test_eq(run('cmd /c echo hi'), 'hi') on Windows
2. def run(cmd, *rest, ignore_ex=False, as_bytes=False, stderr=False) doesn't support redirection in command, does it matter?

jph00:
> 2\. def run(cmd, *rest, ignore_ex=False, as_bytes=False, stderr=False) doesn't support redirection in command, does it matter

No I don't think so - you can use `subprocess.run(..., shell=True)` if you need redirection, or use a more full featured python lib like [sh](https://amoffat.github.io/sh/)

jph00:
> > It's currently marked as a draft. Please mark it as ready to review if you want it merged.
> 
> > It's currently marked as a draft. Please mark it as ready to review if you want it merged.
> 
> @jph00 ， I marked it as draft because I need more time to enable all fastcore support windows.
> Now. all tests are passed in windows.
> image

Excellent - thank you!

Issue #285, opened 2021-01-27
Feature request: Optional arguments for the to-be-tested function in `test_fail`


michaelaye:
The code of `fastcore.test.test_fail` shows that it only calls the provided function without any parameters.
It would be nice to be able to directly provide parameters to the function to test that it fails for some parameters.
In code:

```python
def f(a):
    if a == 5:
        raise ValueError
```
and I would like to force this failure for testing by using something like `test_fail(f, args=(5,))`

Salehbigdeli:
The feature is added and this issue should close :)

PR #286, opened 2021-01-28
Add Optional arguments for the to-be-tested function in `test_fail`

Salehbigdeli:
Adds feature requested in #285

### DIFF
--- a/fastcore/test.py
+++ b/fastcore/test.py
@@ -10,9 +10,10 @@ from collections import Counter
 from contextlib import redirect_stdout
 
 # Cell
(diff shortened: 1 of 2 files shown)


hamelsmu (changes requested):
You need to add associated documentation and tests for the new features being added here

hamelsmu:
@jph00 is this a feature we want to add?  I know that with every new feature there is associated maintenance cost and it seems like you use the anonymous functions everywhere else where you want to run tests like this?  WDYT?

jph00:
LGTM :)

PR #287, opened 2021-01-29
add is_in_ipython

mszhanyi:
Add a function to check if the code is runing in ipython


jph00:
fastcore already provides an `IN_PYTHON` variable you can use :)

PR #289, opened 2021-02-02
update parallel

mszhanyi:
1. As dataloader, set number of worker as 0 in Notebook and multiple workers works in scripts
2. override system default encoding to UTF-8 on Windows


mszhanyi:
@jph00 https://github.com/fastai/fastai/pull/3198 depends on this.

nbs/03_xtras.ipynb
```diff
@@ -41,6 +41,17 @@
     "from time import sleep"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "if sys.platform == \"win32\":\n",
```
jph00, line 50: Looks like this cell isn't exported, so it's not doing anything.

Although I don't think this is a great solution. It's both relying on a private python implementation detail, and it's also changing the behavior of python outside of use of the library.

jph00 (changes requested):
Just one request here

nbs/03_xtras.ipynb
mszhanyi: @jph00 Thank your comments.
similar change made in fastai too. https://github.com/fastai/fastai/pull/3203

PR #294, opened 2021-02-23
parallel in win notebook

mszhanyi:
It is  very tricky to run multiprocessing in notebook on Windows.
First, The multiprocessing can only be executed in [Top-level script environment](https://docs.python.org/3/library/__main__.html), that is in the scope of `if __name__ == "__main__":`. 
Otherwise, notebook will hang due to reimporting modules.

Second, all functions, objects used in the subprocesses must be imported from separate module, not from the main module. Because the main module in newly spawned process in notebook is a empty module. If something used in subprocess but is defined in main module, it will throw `AtrributeError: Can't get attribute ....`

Thirdly,  we have to import libraries in the function. otherwise, we'll get an exception that `module isn't defined`. https://github.com/mszhanyi/fastcore/blob/e1536a5dddbde477e84dfa57e372e4f481353087/fastcore/parallel.py#L128

Lastly, lamda function couldn't be passed to subprocess.

This PR is to provide a example on how to run fastcore.parallel in notebook on Windows.
As for dataloader, It is much harder because there are multiple functions and objects send to workers. code would become very ugly.

@jph00 @maxluk


jph00:
Thanks @mszhanyi - sorry hadn't noticed this wasn't merged yet

renato145:
I noticed that now `parallel` fails with partial functions.
`parallelable` checks `f.__module__` which partial functions don't have.

jph00:
Thanks @renato145 . @mszhanyi is this something you might be able to resolve? I guess if it's partial, or for whatever reason `__module__` isn't available, we should just assume it's not Windows.

PR #305, opened 2021-04-12
windows compatibility for run

mszhanyi:
It needs to add `cmd /c` on Windows for some commands, such as `conda info --root`
For convenience, same_in_win is added for those commands.

@jph00


mszhanyi:
@jph00 I doubt my tests always failed in CI because Anaconda or miniconda isn't installed in the CI machine. Could we install it in CI?

PR #326, opened 2021-06-08
add yaml load to Path

hamelsmu:


jph00:
I'd rather not import yaml in this file - thanks anyway for the PR!

PR #329, opened 2021-06-25
Add file_extract

amaarora:
Move file_extract from fastai.data.external to fastcore.xtras and also add option for xz and bz2 tarfile types


jph00:
It turns out that the python standard lib already handles this for us :) https://docs.python.org/3/library/shutil.html#archiving-operations

PR #339, opened 2021-08-04
Make join_path_file() check if the same extension in file is passed as ext and remove it

4ndr3aR:
Now join_path_file() checks if the same extension passed as ext is already in the file string and removes it. (Made this change at the n-th "no such file or directory <blah.pth.pth>" doing a learn.load() -- with n very big.)


jph00:
I'd rather not make this change - but thanks for contribution!

PR #352, opened 2021-08-12
fixes #351 - L None-ness check infinite recursion

### LINKED ISSUE #351: non-idiomatic L None-ness check infinite recursion
A code snippet like this will fail due to infinite recursion.

```
    some_var = L(['a', 'b'])
    assert some_var != None
```

This results in:

```
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
<ipython-input-105-b7b494b72a34> in <module>
      7 # non-idiomatic None-ness check - avoid infinite recursion
      8 some_var = L(['a', 'b'])
----> 9 assert some_var != None, "L != None"

<ipython-input-104-4baf48a81654> in __eq__(self, b)
     31     def __eq__(self,b):
     32         #if b is None: return False
---> 33         if risinstance('ndarray', b): return array_equal(b, self)
     34         if isinstance(b, (str,dict)): return False
     35         return all_equal(b,self)

~/fast/fastcore/nbs/fastcore/imports.py in array_equal(a, b)
     50     if hasattr(a, '__array__'): a = a.__array__()
     51     if hasattr(b, '__array__'): b = b.__array__()
---> 52     return (a==b).all()
     53 
     54 def df_equal(a,b): return a.equals(b) if isinstance_str(a, 'NDFrame') else b.equals(a)

... last 2 frames repeated, from the frame below ...

<ipython-input-104-4baf48a81654> in __eq__(self, b)
     31     def __eq__(self,b):
     32         #if b is None: return False
---> 33         if risinstance('ndarray', b): return array_equal(b, self)
     34         if isinstance(b, (str,dict)): return False
     35         return all_equal(b,self)

RecursionError: maximum recursion depth exceeded
```

It is easy enough to work around by changing to:

```
    assert some_var is not None
```

But the `!=` approach is also valid Python (to the best of my understanding)
And, in environments with an out-of-memory-killer (like containers),  the process might vanish unexpectedly with no error message.

I've learned my lesson and don't need the change. I am just reporting this in case you consider it a usability improvement for others.

I have a PR ready to go with a fix and a corresponding test in hidden cell.

MarkGass:
See issue #351  which includes the repeating traceback.

I have included a hidden cell test to illustrate the problem. 

The fix passed nbdev_test_nbs for fastcore and fastai.

Sorry. I see the undesirable line number diffs. I am not sure if that is something the notebook cleanup tools should have fixed....

### DIFF
--- a/fastcore/foundation.py
+++ b/fastcore/foundation.py
@@ -127,6 +127,7 @@ class L(GetAttr, CollBase, metaclass=_L_Meta):
             for i,o_ in zip(idx,o): self.items[i] = o_
 
     def __eq__(self,b):
+        if b is None: return False
         if risinstance('ndarray', b): return array_equal(b, self)
         if isinstance(b, (str,dict)): return False
         return all_equal(b,self)
--- a/nbs/02_foundation.ipynb
+++ b/nbs/02_foundation.ipynb
@@ -575,6 +575,7 @@
     "            for i,o_ in zip(idx,o): self.items[i] = o_\n",
     "\n",
     "    def __eq__(self,b): \n",
+    "        if b is None: return False\n",
     "        if risinstance('ndarray', b): return array_equal(b, self)\n",
     "        if isinstance(b, (str,dict)): return False\n",
     "        return all_equal(b,self)\n",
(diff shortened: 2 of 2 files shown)


jph00:
Many thanks. There's no way to avoid the line number diffs - the links to the source code do indeed need to change like this.

Issue #353, opened 2021-08-16
untar_dir puts temporary files in $CWD


mdub:
[`untar_dir` unpacks archives into a temporary directory _within the current working directory_](https://github.com/fastai/fastcore/blob/9806f0034bed52bdcdef4fffb353f01f4eadf4db/fastcore/xtras.py#L178).

This became an issue for me when running "fastbook" notebooks within a [VSCode devcontainer](https://code.visualstudio.com/docs/remote/containers).  The notebook working directory (VSCode "workspace" dir) was on a volume shared (via a Docker bind-mount) with the host (Mac), which is known to add significant overhead for disk IO, and as a result unpacking the IMDB dataset took many minutes. By temporarily changing directory to `/tmp`, I was able to complete the `untar_dir` in under 20 seconds.

AFAICS the location of the "unpacking" area is arbitrary, especially since source `fname` and target `dest` may be elsewhere on the filesystem.  As such, using `dir = '.'` seems unnecessary; could that argument to `TemporaryDirectory` be removed, so that it used the default tempdir?

PR #358, opened 2021-08-22
Use default TemporaryDirectory in "untar_dir"

mdub:
Prevents issues when current working directory is unwritable, or inappropriate for large amounts of data.

Fixes https://github.com/fastai/fastcore/issues/353.

I have _not_ written a test - figuring this is a trivial change, and doesn't alter functionality - but can (attempt to) do so if requested.


jph00:
Thanks

PR #359, opened 2021-08-25
improve documentation for using_attr

k0ala:


k0ala:
@jph00 is there something I need to do here?

jph00:
Thanks @k0ala . It's a good idea to at-mention me so I get notified. Otherwise it can take some time until I next check out the PR queue.

Issue #360, opened 2021-08-25
`pyproject.toml` support?


TeoZosa:
I just started playing around with `nbdev` and I've really been enjoying the experience so far! Thanks to everyone for all their hard work!

# Proposal

With that being said, **would the ability to alternatively specify configuration values in a `pyproject.toml` file be a useful feature**? 
- I know that fastai projects all build on the `nbdev_template` which, in turn, uses a `settings.ini` file, so this wouldn't provide any new functionality, per-se, but it may be helpful for users who already use a `pyproject.toml` file for their other tooling configurations (e.g., Poetry, etc.).

## Pros: 
1. [PEP-518](https://www.python.org/dev/peps/pep-0518/) compatibility
    - As a consequence of [PEP-518](https://www.python.org/dev/peps/pep-0518/), many tools require/provide support for specifying configuration values in a `pyproject.toml` file, and users are therefore increasingly likely to already use one in their project.

## Cons:
1. NO variable substitution; upstream callers must interpolate variables themselves
    - For example: "`%(user)s`"
2. Requires an extra dependency to load/save `.toml` files
    - Viz. `tomlkit` (which preserves formatting, as opposed to the `toml` library).

## Considerations:
1. Section naming (`[tool.fastai]` was used for the example PR).
2. Whether or not to also implement this functionality in `setup.py`
    - Since `settings.ini` is also used in `setup.py`, we could extend this same functionality to `setup.py` (this would be a super simple follow-up PR if the team decides this is the way to go).

## Related:
1. **(potentially worthwhile to mention, but out of scope)**: [PEP-517](https://www.python.org/dev/peps/pep-0517/) support
    - While `setuptools` continues to be the project build backend, code from `setup.py` could be migrated to `pyproject.toml` and `setup.cfg` (see: [setuptools 57.4.0 documentation: Build System Support](https://setuptools.readthedocs.io/en/latest/build_meta.html#build-system-support)).

Thanks!

jph00:
Thanks for the thoughtful issue. I'm open to this, although I'll need to look into it further. The additional dep is an issue though.

I'd rather support just one format. Perhaps toml would be something for us to consider in nbdev2...

TeoZosa:
My pleasure and sounds good! I'd love to contribute in some way when nbdev2 comes around 😄

PR #361, opened 2021-08-25
WIP: add `pyproject.toml` support

TeoZosa:
See: #360


jph00:
@seeM FYI this is a draft of something that uses toml instead of configparser.

Just noting this when we/if consider adding optional pyproject.toml support to nbdev2 later.

PR #375, opened 2021-10-29
add flatten collection func

tcapelle:
Adding a simple function to flatten nested collections:
```python
tuple(flatten(((1,2,3), 4, (5, (6,7)))))
> (1, 2, 3, 4, 5, 6, 7)
```


jph00:
Thanks for this - I ended up doing this manually to help reduce some dupe code

tcapelle:
thanks, this is needed to fix the `MixedPrecision` Callback

PR #378, opened 2021-11-02
Add ismin_version & notmax_version

warner-benjamin:
Adds version checks with packaging.version. For resolving issues with current fastai torch version checking brought up in fastai/fastai#3511.


jph00:
I don't think this is worth having a fastcore function for - and certainly not worth having an export dep in `imports`, which would mean a whole extra module gets parsed everytime anything from fastcore is used. Let's just have a `min_torch_ver` function or something in fastai.

PR #387, opened 2022-01-26
Use docments

muellerzr:
Updates to use docments, cc @jph00


muellerzr:
Mostly whitespace funks. 08_script has the fun

muellerzr:
@jph00 good to go

PR #411, opened 2022-05-19
Add dataclass to docments

MarkB2:
`fastcore.docments.docments` can't show dataclass definition. It tryes to find this class constructor, but it is unaccessible, so it throws exception. 
```python
from dataclasses import dataclass

@dataclass
class Person:
    name:str # The name of the person
    age:int # The age of the person
    weight:float # The weight of the person

docments(Person)
```
Output
```
{ 
    'age': 'The age of the person',
    'name': 'The name of the person',
    'return': None,
    'weight': 'The weight of the person'
}
```
The only caveat is this output would be produced is notebook already converted into script. `docments` of dataclass defined online in notebook couldn't return parameters' comments.
cc @jph00


jph00:
@MarkB2 are you planning to resubmit this, or did you decide against implementing this?

MarkB2:
I am suddenly ruin my system. It will resubmit the PR tomorrow I hope.

PR #413, opened 2022-05-21
Add dataclass to docments

MarkB2:
`fastcore.docments.docments` can't show dataclass definition. It tryes to find this class constructor, but it is unaccessible, so it throws exception. 
```python
from dataclasses import dataclass

@dataclass
class Person:
    name:str # The name of the person
    age:int # The age of the person
    weight:float # The weight of the person

docments(Person)
```
Output
```
{ 
    'age': 'The age of the person',
    'name': 'The name of the person',
    'return': None,
    'weight': 'The weight of the person'
}
```
The only caveat is this output would be produced if notebook already converted into script. `docments` of dataclass defined online in notebook couldn't return parameters' comments.
cc @jph00


jph00:
Thanks @MarkB2 ! If you have a moment can you take a look at the failing test in CI? Looks like you might be using a stdlib feature that's not in py37.

MarkB2:
@jph00, thanks. It's done.

Pugio:
Hi @MarkB2 , could you explain why this is set to only work if the dataclass comes from an imported module?

```python
def get_dataclass_source(s):
    "Get source code for dataclass `s`"
    return getsource(s) if not getattr(s, "__module__") == '__main__' else ""
```

I have a dataclass in my notebook which isn't getting any documentation because it's in my main notebook. I'm not sure of the intent behind this `if` condition, but I don't want to submit a PR until I understand why this is here.

Thank you.

MarkB2:
In Jupyter notebook a new dataclass creates as built in one, meaning it has no source. Thus trying to get class declaration fails.
```python
from inspect import getsource
from dataclasses import dataclass

@dataclass
class MyClass:
    ' my awesome dataclass'
    first: int = 5 # first variable
    second: str = 'Hello' # second variable

getsource(MyClass)
```
throws exception
```python
TypeError: <class '__main__.MyClass'> is a built-in class
```
The other way is to get is through python AST, but you are loosing parameters annotaitions
```python
MyClass??
# Jupyter outputs
# Init signature: MyClass(first: int = 5, second: str = 'Hello') -> None
# Docstring:      my awesome dataclass
# Type:           type
# Subclasses:     
```
For regular class this annotations recorded in class constructor, but dataclass doesn't have (an accessible) one.
I didn't find a better solution. Module creation solves the above problems, so I thought it could be OK.

Pugio:
Thank you, that's very helpful! My actual intent is to generate useful docs from Pydantic models, which seem to act similarly to dataclasses in some respects (ie. no accessible/useful `__init__()` by default). I guess I'll have to put these in modules.

PR #414, opened 2022-05-30
refactor `_TypeDict`

seeM:
#### Context

I'm in the process of deep-diving fastai (specifically the tabular sections) and fastcore while on a sabbatical. I'm a big fan so far - it's so user-friendly, and the programming style is reminding me why I got into programming to begin with 😄! While interacting with the notebooks I'm finding tiny opportunities to clean up the implementation and docs.

I hope this is seen as a positive contribution rather than bike shedding. If it doesn't look valuable, please let me know. I'd be more than happy to change my approach and try something else 😃

For interest sake, this PR is a small part of a broader clean up of the dispatch notebook available in my forked master branch (view the full diff [here](https://github.com/fastai/fastcore/compare/master...seeM:master) and the updated notebook [here](https://github.com/seeM/fastcore/blob/master/nbs/04_dispatch.ipynb)). That branch includes more work along these lines, as well as a restructuring of the docs that I found helpful in developing my understanding of the library. I'd be very happy to continue carving out PRs like this from that branch

#### Description of changes

_(commit messages copied as-is)_

Make it more dict-like with fewer lines where appropriate:

- Rename `add` to `__setitem__`
- Add `setdefault` which cleans up `TypeDispatch.add`

Also renamed args for consistency: vars referring to keys are named `t`
since they will always be types, and vars referring to values are named
`v` (or `vs`) because they may be functions or nested `_TypeDict`s.

### DIFF
--- a/fastcore/dispatch.py
+++ b/fastcore/dispatch.py
@@ -24,7 +24,7 @@ def lenient_issubclass(cls, types):
 
 # Cell
 def sorted_topologically(iterable, *, cmp=operator.lt, reverse=False):
-    "Return a new list containing all items from the iterable sorted topologically"
+    "Return a new list containing all items from `iterable` sorted topologically."
     l,res = L(list(iterable)),[]
     for _ in range(len(l)):
         t = l.reduce(lambda x,y: y if cmp(y,x) else x)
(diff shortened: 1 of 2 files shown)


seeM:
Closing, will focus on #415 instead. cc @jph00

jph00:
> Closing, will focus on #415 instead. cc @jph00

We might want to reopen this if we end up feeling like plum isn't the best approach.

PR #417, opened 2022-06-15
Add a `nested_callable` and `getcallable` util

muellerzr:
# `nested_callable` + `getcallable`

## What does this add?

This PR adds a small util func of `nested_callable`. I was noticing this pattern being repeated a lot in fastai code:
```python
someresult = getattr(obj, name, noop)(arg_a, arg_b, arg_c)
```

This util simplifies that code with a shorthand of `nested_callable` for nested attributes, and `getcallable` for non-nested attributes.

The new version looks like so:
```python
someresult = nested_callable(obj, name)(arg_a, arg_b, arg_c)
```
(And with getcallable):
```python
someresult = getcallable(obj, name)(arg_a, arg_b, arg_c)
```

cc @hamelsmu @jph00


hamelsmu:
LGTM

PR #421, opened 2022-06-27
use py310 style union annotations

seeM:
This changes `eval_types` to parse `|`-separated string annotations to `Union`s instead of tuples, which (1) fixes `@patch` and `@typedispatch` when using py310 style union annotations in Python 3.10 without importing future annotations, and (2) is required by https://github.com/fastai/fastcore/pull/415. It is still compatible with the tuple style.

It also changes all tuple style annotations to py310 style. I found these using [ripgrep](https://github.com/BurntSushi/ripgrep) with `rg --multiline -t jupyter -t py '\([^\(\)]*:\s*\(.*'`.

cc @hamelsmu @jph00


seeM:
@jph00 this is ready for review. This changes `eval_types` to parse `|`-separated string annotations to `Union`s instead of tuples, which (1) fixes `@patch` and `@typedispatch` when using py310 style union annotations in Python 3.10 without importing future annotations, and (2) is required by https://github.com/fastai/fastcore/pull/415. It is still compatible with the tuple style.

It also changes all tuple style annotations to py310 style. I found these using [ripgrep](https://github.com/BurntSushi/ripgrep) with `rg --multiline -t jupyter -t py '\([^\(\)]*:\s*\(.*'`.

jph00:
Thank you!

PR #422, opened 2022-06-30
fix parallel on macos

seeM:
This fixes parallel on MacOS (in my case, specifically for `nbprocess_test`) by only switching to `fork` start mode if in a notebook. Tests pass on fastcore, nbdev, and fastai.

#### Details

The C code (`matplotlib.backends._macosx`) called when enabling matplotlib on an interactive shell within a `ProcessPoolExecutor` using the `fork` start mode errors with:

```
objc[50313]: +[NSMutableString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

However, it works with the `forkserver` start mode, which was recommended in [this discussion](https://bugs.python.org/issue33725). I believe we originally switched to `fork` on MacOS because MacOS' default `forkserver` [doesn't work in IPython](https://stackoverflow.com/a/65001152) - that isn't needed on Linux because Linux already defaults to `fork`.

cc @hamelsmu @jph00

### DIFF
--- a/fastcore/parallel.py
+++ b/fastcore/parallel.py
@@ -10,14 +10,12 @@ from .basics import *
 from .xtras import *
 from functools import wraps
 
-# from contextlib import contextmanager,ExitStack
 from multiprocessing import Process, Queue
 import concurrent.futures,time
 from multiprocessing import Manager, set_start_method
 from threading import Thread
-
 try:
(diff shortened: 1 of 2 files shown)


jph00:
Thanks @seeM !

PR #424, opened 2022-07-01
refactor transform

seeM:
I've split this out from #415 to make that easier to review when it's ready. Minor clean up of `fastcore.transform`, and added two tests corresponding to previously untested fastai use-cases.

cc @hamelsmu @jph00


hamelsmu:
Awesome work

PR #427, opened 2022-07-08
adds mkdir

lgvaz:
Adds `mkdir` function as described in discord [here](https://discord.com/channels/689892369998676007/689892370002870284/993544265093357749).

I feel like my tests might be a bit convoluted? Please feel free to suggest changes :)

**NOTE:** The only file I really modified was `03_xtras.ipynb`, I guess all the other files were changed by nbdev hooks?


lgvaz:
Hmm, some permission error on the CI

jph00:
Yeah you can't delete a directory you're using in Windows. And your `with` block means you're using it.

Change your code so you don't use `with`, but instead use a `finally` block to clean up.

(PS: be sure to at-mention me on PRs so I see them)

lgvaz:
Heh @jph00, I don't know how to do this 😅

jph00:
Hey no worries. You're still using a `with` block for the directory. Your commit only changed the file.

jph00:
Feel free to close the PR if you don't feel like debugging this -- it's not a big deal either way :)

lgvaz:
No no, don't get me wrong, I really wanna do this, I just had spent and hour trying to understand the error and was not getting nowhere so I asked for help, but in hindsight "I don't know how to do this" is one of the worse ways for asking help, sorry about that.

Anyways, the directory is using `with`!!! I completely overlooked that and tunel visioned on the file, I'll try it again now :)

lgvaz:
Yeeeee @jph00 had to be a bit hacky but it worked! I had to manually close the file and the directory for windows to be happy about deleting it hahahah. Let me know if you think there's a better way to do this, maybe again I just focused too much into using the temporary dir/files and missed an simpler approach.

Btw, should I tag you in every update in the PR or just once at the beggining?

jph00:
This is using a private method, which isn't a good idea.

I think you're making this over complicated frankly. There are examples on stack overflow that'll show you how to get a temporary file name without actually creating anything. Afaict you should be fine to use "with" for the file, just not for the directory.

You only need to tag me at the start of a PR

lgvaz:
Hi! Just want to say I have not forgotten about this. I just got my hands on a windows laptop so that will make it much easier to figure this out!

lgvaz:
Hi Jeremy! It's been a while but I finally figured it out. As you suggested, I was overcomplicating: I don't need to create a temporary file inside a temporary directory, I can just create a normal file!!

The CI on windows is currently disabled, but I tested this on a windows machine and it works!

nbs/.gitattributes
```diff
@@ -1 +1,2 @@
-*.ipynb merge=nbdev-merge
```
jph00, line 1: Can you revert your change to this file please? Looks like you've got an old version of nbdev hooks installed that's created this change.

nbs/03_xtras.ipynb
```diff
@@ -318,6 +318,40 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#|export\n",
+    "def mkdir(path, exist_ok=False, parents=False, overwrite=False, **kwargs):\n",
+    "    \"Creates and returns a directory defined by `path`, optionally removing previous existing directory if `overwrite` is `True`\"\n",
+    "    path = Path(path)\n",
```
jph00, line 330: You should add `import shutil` here. Currently this import is only in a non-exported cell, so this won't work outside this notebook.

jph00 (changes requested):
Looking very close now!

nbs/.gitattributes
lgvaz: done :)

nbs/03_xtras.ipynb
lgvaz: done! it should be inside the function right?

lgvaz:
I see there's a conflict now. Is it because the cell number of my added function has changed compared to master?

lgvaz:
Done! :)

jph00:
Good stuff!

PR #428, opened 2022-07-11
add nested parser support to `fastcore.script`

seeM:
There are some limitations, see the docs on `--`. Also some behaviour I don't understand, like `nbscript --script_name=update update-union-annots.ipynb -- -h` works as intended, but `nbscript update-union-annots.ipynb --script_name=update -- -h` doesn't. `--` is ignored in the first example but parsed in the second.

Example use-case - run notebooks as scripts:

```python
from fastcore.script import *
from nbprocess.export import nb_export
from nbprocess.processors import _hide_dirs
from pathlib import Path
from runpy import run_path
from tempfile import TemporaryDirectory

def keep_code(cell):
    "Remove cells that aren't exported or hidden (opposite of `rm_export`)"
    if not (cell.directives_ and (cell.directives_.keys() & _hide_dirs) and cell.cell_type == 'code'): del(cell['source'])

@call_parse(nested=True)
def nbscript(nb_path:str, # Notebook path to execute 
             script_name:str=None): # Name of the script used in the help message (default: `nb_path`  with special chars replaced by `_`)
    "Run notebooks as scripts"
    nb_path = Path(nb_path)
    script_name = script_name or nb_path.stem.replace('-','_')
    with TemporaryDirectory() as td:
        lib_path = Path(td)/script_name
        nb_export(nb_path, lib_path)
        run_path(str(lib_path), run_name='__main__')
```

### DIFF
--- a/fastcore/script.py
+++ b/fastcore/script.py
@@ -95,8 +95,10 @@ def args_from_prog(func, prog):
 SCRIPT_INFO = SimpleNamespace(func=None)
 
 # Cell
-def call_parse(func):
+def call_parse(func=None, nested=False):
     "Decorator to create a simple CLI from `func` using `anno_parser`"
+    if func is None: return functools.partial(call_parse, nested=nested)
+
     mod = inspect.getmodule(inspect.currentframe().f_back)
     if not mod: return func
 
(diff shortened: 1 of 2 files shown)


seeM:
@jph00 I consider it experimental. Could still be merged if you're happy

jph00:
lgtm!

PR #431, opened 2022-07-15
Export _get_property_name from 06_docments.ipynb

warner-benjamin:
Export `_get_property_name` from 06_docments.ipynb so `get_name` can call it.


warner-benjamin:
@jph00 Didn't realize I need to manually tag you in this repo for a PR. This one should be a quick review though 😄

jph00:
Thank you!

PR #436, opened 2022-07-18
add context manager for changing directories

hamelsmu:
NB: this will be included in Python eventually https://github.com/python/cpython/pull/28271

Going to use this feature for running notebooks with execnb in the right directory


hamelsmu:
This already exists 🤦🏽

PR #443, opened 2022-07-28
fix ci regex

seeM:
cc @hamelsmu @jph00


hamelsmu:
Re running the integration tests here (I've merged the other related PRs)

jph00:
Yay all passing!

PR #449, opened 2022-08-02
Fix scripts not executing

renato145:
I noticed my scripts using `@call_parse` where not working because when checking if the script was being called from main there was some quotes missing in the code:
`if getattr(mod, __name__, '') =="__main__": ...`
instead of:
`if getattr(mod, '__name__', '') =="__main__": ...`


ew-meetup:
Interesting. I think this caused a hack I had to get around this issue to regress.

Mentioning just in case someone else was using fastcore's `call_parse` with PySpark (or a similar indirect runner). Having this at the bottom of each script shouldn't be necessary anymore:

```
#|export
#|eval: false

# This is a hack because spark-submit doesn't actually play nice with fastcore's call_parse out of the box
# https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/PythonRunner.scala#L75
# https://github.com/fastai/fastcore/blob/1dc01246e4553487157099221f0085676919d746/nbs/08_script.ipynb

if __name__ == "__main__":
  locals().get("my_method")()

```

yegeniy:
Interesting. Thanks @renato145. 
I think this caused a hack I had to get around this issue to regress.

Mentioning that only in case someone else was using fastcore's `call_parse` with PySpark (or a similar indirect runner). Having something like this at the bottom of each script shouldn't be necessary anymore:

```
#|export
#|eval: false

# This is a hack because spark-submit doesn't actually play nice with fastcore's call_parse out of the box
# https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/PythonRunner.scala#L75
# https://github.com/fastai/fastcore/blob/1dc01246e4553487157099221f0085676919d746/nbs/08_script.ipynb

if __name__ == "__main__":
  locals().get("my_method")()

```

Issue #445, opened 2022-07-29
AttrDict


erlebach:
Consider: 
a = L(3,4)  # a wrapper around a list. b = a.copy()  # copies as expected and b is of type L
Do the same with a dictionary:
a = AttrDict({'a':3, 'b':4})b = a.copy()
b is a regular dict, and not an AttrDict. Is this expected behavior? I had a bug related to this issue. 

Great library by the way.

jph00:
No not what I'd expect as behavior. Feel free to submit a PR if you have a good solution.

Salehbigdeli:
> No not what I'd expect as behavior. Feel free to submit a PR if you have a good solution.

Just submitted a PR for it!

PR #451, opened 2022-08-04
New method: `AttrDict.copy`, to return AttrDict instead of plain python dict

Salehbigdeli:
Suggested solution to #445


jph00:
Great thanks!

PR #456, opened 2022-08-06
layered `Config` with `extra_files`; `Config` `repr`

seeM:
To be used in `nbdev` with `fastcore.xdg` for combining user and project settings, cc @hamelsmu @jph00


jph00 (changes requested):
Just one suggestion. Thanks for adding the docs! :D

PR #459, opened 2022-08-08
enable meta.delegates to enforce KEYWORD_ONLY for kwargs from

EmbraceLife:
The details of this PR can be found in this forum [thread](https://forums.fast.ai/t/python-help/97525/22?u=daniel).


fastcore/meta.py
```diff
@@ -120,20 +120,21 @@ def _f(f):
         k = sigd.pop('kwargs')
         s2 = {k:v for k,v in inspect.signature(to_f).parameters.items()
```
jph00, line 121:
```suggestion
        s2 = {k:v.replace(kind=inspect.Parameter.KEYWORD_ONLY) for k,v in inspect.signature(to_f).parameters.items()
```

If you made this change, would that mean you could then remove the line you added?

jph00 (changes requested):
Many thanks for the PR. I've made one suggested change (but haven't tested it -- so take a look and see if it works for you!).

You've accidentally added the whole _docs directory to your PR. Could you please remove it? I've just updated .gitignore so that this shouldn't happen again in the future.

fastcore/meta.py
EmbraceLife: Yes, this change works and my added line can be removed. Thanks Jeremy!

jph00: Great. Just reply here when you've removed the `_docs` folder and we should be ready to go!

EmbraceLife:
I have reverted my first troublesome commit and then only run `nbdev_prepare` this time without running all the cells, nor running `nbdev_docs` and `nbdev_preview`, so that I didn't generate so many differences.

jph00:
Good job!

EmbraceLife:
<p>Thank you Jeremy! This is my first contribution to fastcore, Wow!</p>

PR #460, opened 2022-08-10
Add `fastcore.style`: fast styling for friendly CLIs

seeM:
cc @hamelsmu @jph00


jph00:
I like it! Two suggestions:

1. Inherit from, or (maybe better?) compose, `list`, not `L`, to avoid dep on `foundation`
2. Put this in its own module -- since it'll be used in a lot of CLIs, we want it fast, so no reason to import all of `xtras`

Also, can you check whether Quarto now supports ansi terminal escapes? If it does, we should stop removing them all in processors.py.

seeM:
@jph00 Done, I like your suggestions! In my testing, it doesn't seem like Quarto supports ansi terminal escapes yet.

PR #464, opened 2022-08-14
exclude `passed to {func}` from docments by passing `verbose=False` to `delegates`

seeM:
cc @hamelsmu @jph00


jph00:
There isn't really enough docs here (and there's no PR description) for me to know what's going on, so I don't have anything intelligent to say about this, sorry @seeM ! Maybe you could add some more docs to the notebook to explain this?

seeM:
@jph00 Sorry about that! I want to use `delegates` to a private function without affecting `show_doc`. Specifically, I wanted to remove all of the "passed to `apply_defaults`" in [`nbdev_create_config`](https://nbdev.fast.ai/read.html#nbdev_create_config)'s docs:

image

The approach taken here was to add a `verbose` argument to `delegates`, which is set in a dict at `from_func.__delopts__`. Later, `docments` uses that value to decide whether to include "passed to `to.__name__`" or not.

I documented the arg on `delegates` but in hindsight `# Include \`to\` in docments?` isn't very informative. I also added a small example to the docments nb but it might've gotten lost among the other smaller doc changes I made there - or does it need to be made more informative too?

image

As always, I'm very happy to change/improve/scrap any of this :) don't hold back!

jph00:
OK great all clear! :)

Issue #465, opened 2022-08-14
conda install command in docs doesn't work


Jorgvt:
Hi!

Long time lurker here. I finally decided to try `fastcore`, but when following the docs it's said to install the library by doing `conda install fastcore`, but I couldn't do it because it couldn't be found. I just fixed it by doing `conda install fastcore -c fastai`.

Im not sure if it's something on my end that is not configured properly on `conda`, but just wanted to make sure you knew about it. If it's only a docs typo, I would be more than happy to try and fix it (never done a PR, but would love to try).

Thank you for your awesome work!

PR #469, opened 2022-08-20
conda install command in docs doesn't work

Jorgvt:
Hello!

I decided to venture into creating a PR regarding #465 . I don't know if it's considered rude to open a PR straight away (I've never done this before), but after a couple of days of opening the issue, I thought I could give it a try because it seemed easy to fix.

The only thing I have done is adding fastai's channel to the `conda` install command and rebuild the docs so that it's reflected in the `README.md` file.

Best regards!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@ and bringing ideas from NumPy over to Python’s `list` type.
 
 ## Getting started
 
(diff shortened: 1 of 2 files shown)


jph00:
No not rude at all - quite the opposite! :D Many thanks for your contribution.

PR #480, opened 2022-09-05
add body to error message

### LINKED ISSUE #479: Add more information upon HTTP request failures
> one thing that would be really awesome, if you have a chance, if to make fastcore include the actual content of the HTTP body when there's an exception

This might help with troubleshooting https://github.com/fastai/nbdev/issues/944 and https://github.com/fastai/nbdev/issues/916

hamelsmu:
Closes #479 

cc: @seeM @jph00

### DIFF
--- a/fastcore/net.py
+++ b/fastcore/net.py
@@ -17,8 +17,8 @@ __all__ = ['url_default_headers', 'ExceptionsHTTP', 'urlquote', 'urlwrap', 'HTTP
 # %% ../nbs/03b_net.ipynb 1
 from .utils import *
 from .parallel import *
+
 from functools import wraps
-from pprint import pformat
 import json,urllib,contextlib
 import socket,urllib.request,http,urllib
 from contextlib import contextmanager,ExitStack
@@ -106,9 +106,8 @@ def urlopen(url, data=None, headers=None, timeout=None, **kwargs):
         if not isinstance(data, bytes): data = data.encode('ascii')
     try: return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
     except HTTPError as e: 
-        e.msg += f"\n====Error Body====\n{pformat(json.loads(e.read()))}"
-        raise e
-    except Exception as e: raise e
+        e.msg += f"\n====Error Body====\n{e.read().decode(errors='ignore')}"
+        raise
 
 # %% ../nbs/03b_net.ipynb 20
 def urlread(url, data=None, headers=None, decode=True, return_json=False, return_headers=False, timeout=None, **kwargs):
--- a/fastcore/net.py
+++ b/fastcore/net.py
@@ -110,26 +110,26 @@ def urlopen(url, data=None, headers=None, timeout=None, **kwargs):
         raise e
     except Exception as e: raise e
 
-# %% ../nbs/03b_net.ipynb 18
+# %% ../nbs/03b_net.ipynb 20
 def urlread(url, data=None, headers=None, decode=True, return_json=False, return_headers=False, timeout=None, **kwargs):
     "Retrieve `url`, using `data` dict or `kwargs` to `POST` if present"
     try:
         with urlopen(url, data=data, headers=headers, timeout=timeout, **kwargs) as u: res,hdrs = u.read(),u.headers
     except HTTPError as e:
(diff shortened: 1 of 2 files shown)


fastcore/net.py
```diff
@@ -104,28 +104,32 @@ def urlopen(url, data=None, headers=None, timeout=None, **kwargs):
     if data is not None:
         if not isinstance(data, (str,bytes)): data = urlencode(data)
         if not isinstance(data, bytes): data = data.encode('ascii')
-    return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
+    try: return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
+    except HTTPError as e: 
+        e.msg += f"\n====Error Body====\n{pformat(json.loads(e.read()))}"
+        raise e
+    except Exception as e: raise e
```
jph00, line 111: I think you can remove this line

fastcore/net.py
```diff
@@ -104,28 +104,32 @@ def urlopen(url, data=None, headers=None, timeout=None, **kwargs):
     if data is not None:
         if not isinstance(data, (str,bytes)): data = urlencode(data)
         if not isinstance(data, bytes): data = data.encode('ascii')
-    return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
+    try: return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
+    except HTTPError as e: 
+        e.msg += f"\n====Error Body====\n{pformat(json.loads(e.read()))}"
+        raise e
```
jph00, line 110: Should there be a "from None" or something at the end here to avoid a complex stack trace?

fastcore/net.py
```diff
@@ -104,28 +104,32 @@ def urlopen(url, data=None, headers=None, timeout=None, **kwargs):
     if data is not None:
         if not isinstance(data, (str,bytes)): data = urlencode(data)
         if not isinstance(data, bytes): data = data.encode('ascii')
-    return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
+    try: return urlopener().open(urlwrap(url, data=data, headers=headers), timeout=timeout)
+    except HTTPError as e: 
+        e.msg += f"\n====Error Body====\n{pformat(json.loads(e.read()))}"
```
jph00, line 109: The error won't always be json will it?

fastcore/net.py
```diff
@@ -104,28 +104,32 @@ def urlopen(url, data=None, headers=None, timeout=None, **kwargs):
...
-# %% ../nbs/03b_net.ipynb 18
+# %% ../nbs/03b_net.ipynb 20
 def urlread(url, data=None, headers=None, decode=True, return_json=False, return_headers=False, timeout=None, **kwargs):
     "Retrieve `url`, using `data` dict or `kwargs` to `POST` if present"
     try:
         with urlopen(url, data=data, headers=headers, timeout=timeout, **kwargs) as u: res,hdrs = u.read(),u.headers
     except HTTPError as e:
-        if 400 <= e.code < 500: raise ExceptionsHTTP[e.code](e.url, e.hdrs, e.fp) from None
+        if 400 <= e.code < 500: raise ExceptionsHTTP[e.code](e.url, e.hdrs, e.fp, msg=e.msg) from None
         else: raise
```
jph00, line 120: Actually maybe this line isn't needed either?

jph00 (changes requested):

fastcore/net.py
hamelsmu: I think that is still needed?  What if the the code is not betwen 400 and 500?

fastcore/net.py
hamelsmu: good catch

fastcore/net.py
hamelsmu: Decided that `from None` doesn't do anything here, but changed it to just `raise` instead of `raise e`

fastcore/net.py
hamelsmu: good point

hamelsmu:
@jph00 thanks for the review.  You can take a look when you are ready again, I've made changes as discussed

fastcore/net.py
jph00: Oh right yeah I misread the indentation.

PR #482, opened 2022-09-07
fix: `Config.get` returns `False` instead of `None` for missing `bool` keys

seeM:
This was causing a bug in nbdev since `nbdev.config._apply_defaults` only sets the default value if the key was missing, by checking `cfg.get(k) is None`, which was never true for missing `bool`s

nbdev should probably use `k not in cfg` instead of `cfg.get(k) is None`, but I consider this to also be a bug in fastcore


seeM:
cc @hamelsmu @jph00

PR #486, opened 2022-09-10
fix: error in IPython while handling `HTTP4xxClientError`

seeM:
For example:

```pycon
In [1]: from ghapi.core import *

In [2]: GhApi().repos.get('org', 'not-a-real-repo-123123123')
Unexpected exception formatting exception. Falling back to standard exception
Traceback (most recent call last):
  File "/Users/seem/.pyenv/versions/3.9.13/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3398, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-b91955617868>", line 1, in <cell line: 1>
    GhApi().repos.get('org', 'not-a-real-repo-123123123')
  File "/Users/seem/code/ghapi/ghapi/core.py", line 61, in __call__
    return self.client(self.path, self.verb, headers=headers, route=route_p, query=query_p, data=data_p)
  File "/Users/seem/code/ghapi/ghapi/core.py", line 118, in __call__
    res,self.recv_hdrs = urlsend(path, verb, headers=headers or None, debug=debug, return_headers=True,
  File "/Users/seem/code/fastcore/fastcore/net.py", line 217, in urlsend
    return urlread(req, return_json=return_json, return_headers=return_headers)
  File "/Users/seem/code/fastcore/fastcore/net.py", line 118, in urlread
    if 400 <= e.code < 500: raise ExceptionsHTTP[e.code](e.url, e.hdrs, e.fp, msg=e.msg) from None
fastcore.basics.HTTP404NotFoundError: HTTP Error 404: Not Found
====Error Body====
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-a-repository"
}


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/seem/.pyenv/versions/3.9.13/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 1990, in showtraceback
    if hasattr(value, "_render_traceback_"):
  File "/Users/seem/.pyenv/versions/3.9.13/lib/python3.9/tempfile.py", line 472, in __getattr__
    file = self.__dict__['file']
KeyError: 'file'
```

cc @hamelsmu @jph00


fastcore/net.py
```diff
@@ -89,10 +89,15 @@ def urlopener():
     (431,'Header Fields Too Large'),(451,'Legal Reasons')
 )
 
+def _mk_error(nm, code, msg):
+    cls = get_class(nm, 'url', 'hdrs', 'fp', sup=HTTP4xxClientError, msg=msg, code=code)
+    def _init(self, url, hdrs, fp, msg=msg, code=code): super(cls, self).__init__(url, code, msg, hdrs, fp)
+    cls.__init__ = _init
```
seeM, line 95: The crux is to override the `__init__` set by `get_class` to call `super().__init__`.

The `_mk_error` function was needed so that `_init` is included as-is in the closure. Without `_mk_error`, each iteration of the for loop would overwrite the previously defined `_init`.

jph00: This is getting pretty obscure! Is there any simpler way to do this, even if a bit more verbose?

seeM: Do you like having separate classes for each error code?

seeM: I think I'd prefer to use the `urllib.HTTPError` class

jph00: The thing I do want is to be able to catch exceptions based on what kind of error occurred. Is there a way to do that without separate classes?

jph00: Would using `type(name, bases, attributes)` directly be an easier way to create these?

seeM: I think like so:

```python
try: urlopen(...)
except HTTPError as e:
    if e.code != 401: raise
    # ...
```

seeM: > Would using `type(name, bases, attributes)` directly be an easier way to create these?

I thought about it. `type(nm, (HTTP4xxClientError,), {'code': code, 'msg': msg})` works and is concise, but would be a tiny breaking change in the signature, from current:

```python
def __init__(self, url, hdrs, fp, msg=default_msg, code=default_code):
```

to the parent class':

```python
def __init__(self, url, code, msg, hdrs, fp)
```

That might be okay though

jph00: Could you add `__init__` to those attrs to get the init we want?

fastcore/net.py
```diff
@@ -91,7 +91,8 @@ def urlopener():
 
 for code,msg in _httperrors:
     nm = f'HTTP{code}{msg.replace(" ","")}Error'
-    cls = get_class(nm, 'url', 'hdrs', 'fp', sup=HTTP4xxClientError, msg=msg, code=code)
+    def _init(self, url, hdrs, fp, msg=msg, code=code): HTTP4xxClientError.__init__(self, url, code, msg, hdrs, fp)
+    cls = type(nm, (HTTP4xxClientError,), {'__init__':_init})
```
seeM, line 95: @jph00 I think this is much better with your suggestion to use `type`.

Another thing that simplifies it is to directly call `HTTP4xxClientError.__init__` instead of `super(...)`, because:

1.  Using `super` dynamically like this requires explicitly passing the `cls`. And in order to pass a different `cls` each time, we need to define `_init` in a closure to avoid overwriting it with each `cls`.
2. Since `_init` needs a reference to `cls`, we have to first define it with `type(...)` and then patch `cls.__init__`.

This of course has the downside that `HTTP4xxClientError` is hardcoded instead of having `super`'s behaviour.

jph00:
Yay!

fastcore/net.py
seeM: Out of interest, this seems to work too, but I think it's more obscure:

```python
for code,msg in _httperrors:
    class _(HTTP4xxClientError):
        def __init__(self, url, hdrs, fp, msg=msg, code=code): super().__init__(url, code, msg, hdrs, fp)
    _.__name__ = f'HTTP{code}{msg.replace(" ","")}Error'
    _.__qualname__ = _.__qualname__.split('.')[0] + _.__name__
    globals()[nm] = ExceptionsHTTP[code] = _
```

jph00: I guess in that case I'd have a function like:

```python
get _httperrcls(url, code, msg, hdrs, fp):
    class _c(HTTP4xxClientError):
            def __init__(self, url, hdrs, fp, msg=msg, code=code): super().__init__(url, code, msg, hdrs, fp)
     _c.__name__ = f'HTTP{code}{msg.replace(" ","")}Error'
     _c.__qualname__ = _.__qualname__.split('.')[0] + _.__name__
    return _c
```

PR #492, opened 2022-10-13
Remove dash from heading

dleen:
For some reason this dash in the heading causes the TOC in the docs to render incorrectly

@jph00

### DIFF
--- a/nbs/02_foundation.ipynb
+++ b/nbs/02_foundation.ipynb
@@ -2183,7 +2183,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Config -"
+    "## Config"
    ]
   },
   {


dleen:
Unfortunately I wasn't able to figure out why this is happening. I tried creating some notebooks and rendering them directly with quarto and the TOC was generated fine

## Before
before1
before2

## After
after1
after2

jph00:
I might try to figure out why it's broken rather than merging this. Will let you know what I find.

dleen:
Sure, I couldn't figure it out myself but am definitely curious. Although I was confused why there is a dash in the heading in the first place - hence the PR to remove it unless it was intentional?

jph00:
On Thu, Oct 13, 2022, at 10:50 AM, David Leen wrote:
> Sure, I couldn't figure it out myself but am definitely curious. Although I was confused why there is a dash in the heading in the first place - hence the PR to remove it unless it was intentional?
> 

It's intentional. Although now I think about it I'm using it in the wrong way. It actually is used to hide a whole section, but I thought it just hides the heading. So it's entirely my fault!

jph00:
Many thanks!

PR #501, opened 2022-12-09
Update CONTRIBUTING.md

gsganden:
The most important change here is updating `nbdev_install_git_hooks` to `nbdev_install_hooks`, which I believe is necessary for compatibility with the current version of `nbdev`.


jph00:
Many thanks!

PR #506, opened 2022-12-20
remove red lines from image

ddobrinskiy:
clean-up the image by cropping out red lines on the right and bottom

Before: 
Screenshot 2022-12-20 at 15 19 58

After:
Screenshot 2022-12-20 at 15 20 04


jph00:
Thanks :)

PR #507, opened 2022-12-21
extend test_eq_type

ddobrinskiy:
Change the behaviour of `test_eq_type` to fail if comparing tensors with different dtypes

previously this would pass: `test_eq_type(torch.zeros(10), torch.zeros(10, dtype=torch.float64))`

Now it does not


jph00:
Good idea!

jph00:
Oh actually we can't do this because it introduces dependencies on pytorch, numpy, and pandas.

ddobrinskiy:
Oh right, this makes sense.

I did not notice that those packages are part of **dev** dependencies, not main ones.

PR #508, opened 2022-12-21
refactor `00_test.ipynb`

ddobrinskiy:
refactor tests

- add some tests
- update some assert messages
- rewrite some asserts
- all without a change in functionality


jph00:
Many thanks! FYI it's a good idea to at-mention me on PRs so I get notified.

PR #511, opened 2023-01-25
L argfirst bug fix

Isaac-Flath:
@jph00 

Fixed a bug where there was a negate argument that did not have it's behavior implemented.  Added a test for this functionality as well.


jph00:
Many thanks.

Issue #510, opened 2023-01-23
`patch(cls_method=True)` not supporting subclasses correctly


mone27:
I am having an issue with fastcore `patch` and class methods.
If you create a class method in a parent class using `patch(cls_method=True)`  then in subclasses you can still use it but `cls` is wrongly referring to the parent class and not the subclass.
Here there is a little demo of the behavior of the python standard library vs fastcore

```python
# Python standard library
class Parent:
    @classmethod
    def type_cls(cls): return cls

class Child(Parent):
    pass

parent = Parent()
print(parent.type_cls()) # <class '__main__.Parent'>
child = Child()
print(child.type_cls()) # <class '__main__.Child'>

# Fastcore

from fastcore.basics import patch
class FastParent:
    pass

@patch(cls_method=True)
def type_cls(cls: FastParent): return cls

class FastChild(FastParent):
    pass


parent = FastParent()
print(parent.type_cls()) # <class '__main__.FastParent'>
child = FastChild()
print(child.type_cls()) # <class '__main__.FastParent'> while expected is <class '__main__.FastChild'>
```

for comparison normal methods are correctly patched and `self` refers to the subclass

<details>
  <summary> Example with normal method </summary>

```python
# Python standard library
class Parent:
    def type_self(self): return type(self)
    @classmethod
    def type_cls(cls): return cls

class Child(Parent):
    pass

parent = Parent()
print(parent.type_self(), parent.type_cls()) # <class '__main__.Parent'>
child = Child()
print(child.type_self(), child.type_cls()) # <class '__main__.Child'>

# Fastcore

from fastcore.basics import patch
class FastParent:
    pass

@patch
def type_self(self: FastParent): return type(self)
@patch(cls_method=True)
def type_cls(cls: FastParent): return cls

class FastChild(FastParent):
    pass


parent = FastParent()
print(parent.type_self(), parent.type_cls()) # <class '__main__.Parent'>
child = FastChild()
print(child.type_self()) # <class '__main__.Child'>
print(child.type_cls()) # <class '__main__.Parent'> while expected is <class '__main__.Child'>
```
</details>

PR #517, opened 2023-03-02
Bugfix patch classmethod not supporting subclasses correctly

Salehbigdeli:
Fixes: #510


Salehbigdeli:
@jph00 As you told me to mention you :smile:

jph00:
Thank you - sorry for being so slow about this!

PR #518, opened 2023-03-03
Fix minor typo in `patch` documentation

kbressem:
While reading the docs of [`patch_to`](https://fastcore.fast.ai/basics.html#patch_to) and [`patch`](https://fastcore.fast.ai/basics.html#patch) I found a minor typo: 

```# we initilized `t` to a type int = 1``` which I have changed to ```# we initialized `t` to a type int = 1```

```# this sets self.g = 'helllo '```which was changed to ```# this sets self.g = 'hello '```


jph00:
Thanks!

PR #528, opened 2023-05-18
allow docments to parse async function definitions fixes #527

### LINKED ISSUE #527: Allow docments to work with async function definitions
docments does not properly handle async function definitions:

```python
async def add_async(
    # The first operand
    a:int,
    # This is the second of the operands to the *addition* operator.
    # Note that passing a negative value here is the equivalent of the *subtraction* operator.
    b:int,
)->int: # The result is calculated using Python's builtin `+` operator.
    "Add `a` to `b`"
    return a+b
```
> docments(add_async)

```python
{'a': None, 'b': None, 'return': None}
```

kcaylor:
I ran `nbdev_install_hooks` and also `nbdev_clean` against the edited notebook before committing changes. However, it looks like the execution counts are still in the .ipynb file.  

If there's something more/better to do, I can update the request. First time contributing to an nbdev project, so still learning the ropes.

### DIFF
--- a/fastcore/docments.py
+++ b/fastcore/docments.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import re
 from tokenize import tokenize,COMMENT
-from ast import parse,FunctionDef,AnnAssign
+from ast import parse,FunctionDef,AsyncFunctionDef,AnnAssign
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


jph00:
Thank you - sorry for the slow merge!

PR #541, opened 2024-02-07
[Onboarding Experience] Explained how to import fastcore

UmerHA:
Hi all,

I found using `fastcore` a bit tedious because it wasn't clear to me how to import it correctly. I now understand that `from fastcore.all import *` or `from fastcore.SUBMODULE import *` is correct.

This PR adds a 1-line-explainer in the Getting Started section of the docs.

As far as I can tell this is not explained anywhere in the docs yet. This would have saved me a lot of time. :)

### DIFF
--- a/nbs/index.ipynb
+++ b/nbs/index.ipynb
@@ -49,6 +49,8 @@
     "- `fastcore.dispatch`: Multiple dispatch methods\n",
     "- `fastcore.transform`: Pipelines of composed partially reversible transformations\n",
     "\n",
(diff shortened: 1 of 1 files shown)


jph00:
This isn't quite correct @UmerHA . Whilst you *can* import all symbols with a wildcard import (and fastcore is designed to support that well), many people prefer to instead use a module prefix instead, or just import a subset of symbols. This is a python-level issue, not specific to fastcore.

PR #543, opened 2024-04-11
fix: expose timeout to avoid getting blocked indefinitely in certain circumstances

HarikrishnanBalagopal:
Related https://github.com/fastai/ghapi/pull/174


jph00:
Many thanks. This is an nbdev project so the source is the notebook, not the .py file. I'll go ahead and edit it directly.

PR #488, opened 2022-09-28
Enhancement: add ability to have *args in @call_parse

Salehbigdeli:
Fixes: #487

PR #564, opened 2024-05-30
fixes 487

### LINKED ISSUE #487: using `*args` with `@call_parse`
I am trying to the use the `call_parse` decorator with `*args` and I can't seem to figure how to make it work. Here is the function definition,
```python
@call_parse
def timesheet(*args):
  pass
```
When I call the function in the nb, there's no problem,
```
timesheet(2, 'OH', 3, 'CSDP', 3, 'OR')
```
however, while calling the functions from the terminal becomes a problem
```
(base) devengqc@17-CN0065CL-1:~/git/projects/personal$ timesheet --args "2, 'OH', 3, 'CSDP', 3, 'OR'"
usage: timesheet [-h] args
timesheet: error: argument args: invalid _empty value: "2, 'OH', 3, 'CSDP', 3, 'OR'"
```
I've tried other combinations as well in the terminal and they don't seem to work. Is there a way around this? Or am I doing something wrong?

deven367:
This PR is a continuation of the previous PR #488. The PR closes #487 and closes #488.


deven367:
H/T to @Salehbigdeli for fixing the issue. I created a new PR because I could push to the existing PR in #488.

@jph00 can you take a look at this?

jph00:
Thank you! The fastcore CI test can be ignored -- that was my fault and it's fixed now. Can you check the nbdev integration test? Does this PR change the behavior in a user-facing way?

deven367:
@jph00 I don't think the PR changes the behavior in a user facing way.

The integration test fails due to a missing statement in `Returns` in while processing the `DocmentTbl` for the function
```
def _f(a,
        b, #param b
        c  #param c
       ): ...

```

The variable 
```
_exp_res2 = """
|    | **Details** |
| -- | ----------- |
| a |  |
| b | param b |
| c | param c |
"""
```
should be

```
_exp_res2 = """
|    | **Details** |
| -- | ----------- |
| a |  |
| b | param b |
| c | param c |
| **Returns** |  |
"""
```

So the functions that don't have a specific return type, should they have a returns in the `DocmentTbl`?

jph00:
I think it must be this PR that's causing the test failure. Other PRs to fastcore the last few days haven't triggered it, and this PR is changing the anno result for 'returns' to an empty string. Could you possibly take a closer look?

deven367:
Hi @jph00, I tried to make sense of the failing test case, but I can't understand how is anno result and return is being populated :sweat_smile:

PR #565, opened 2024-05-31
Added iter-check to L.__eq__

UmerHA:
Added iter-check to `L.__eq__` as per discord discussion & verified with this fastcore and triton work together.


jph00:
Thank you!

PR #584, opened 2024-07-24
do not list bytes

ababino:
Currenlty, if you do `listify(b'hi')` you get `[104, 105]`. Wouldn't it be better to match how `listify` behaves with strings? If so, this PR changes that. In this implementation, `listify(b'hi')` returns `[b'hi']`


jph00:
Makes perfect sense!

Issue #588, opened 2024-07-31
Interoperability with Django and Jinja using __html__() protocol


collinanderson:
Would be nice to be able to use fasthtml within an existing Django project, as an alternative to Django's html templates.

Django and Jinga already interop with each other using a special __html__() method for communicating whether a string has already been html-escaped.

https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.escape

PR #589, opened 2024-07-31
Interoperability with Django and Jinja using __html__() protocol

collinanderson:
https://github.com/fastai/fastcore/issues/588


jph00:
Great idea!

PR #592, opened 2024-08-04
Remove inline assert on tuple type

pydanny:
As any type can be sent through, including this assertion of type within the method makes it fragile. While inline assertions for types can make sense under the right conditions (aka Django model responses), this isn't one of those occasions.


jph00:
Updated the annotation.

PR #594, opened 2024-08-09
Feat: add an inline flag for to_xml()

phact:
FastHTML is impropertly rendering `<pre>` tags by adding whitespace (https://github.com/AnswerDotAI/fasthtml/issues/235). This PR adds an inline flag to to_xml that can be called to generate xml in a single line.

This is my first time contributing to a fastai project, thanks in advance for your attention! Please let me know if you can think of a better approach for this!

### DIFF
--- a/fastcore/xml.py
+++ b/fastcore/xml.py
@@ -96,13 +96,14 @@ def _to_attr(k,v):
     return f'{k}={qt}{v}{qt}'
 
 # %% ../nbs/11_xml.ipynb
-def to_xml(elm, lvl=0, inline=False):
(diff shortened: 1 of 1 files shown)


jph00:
lgtm! however note that this is an nbdev project, so you actually have to edit the source notebook -- the .py files are auto-generated from that.

jph00:
I've popped it into the notebook and made some minor changes -- will credit you in the release notes.

jph00:
BTW I've also added a `fh_cfg.inline` bool attr for setting this globally in fasthtml.

PR #605, opened 2024-08-14
inline_tags

phact:
related to https://github.com/AnswerDotAI/fasthtml/issues/235


jph00:
I think we've got this at least somewhat sorted now

Issue #608, opened 2024-08-16
Allow urlsend to accept and pass decode flag to urlrequest


radam9:
Hey there,

This issue was already brought up in the following issue https://github.com/fastai/fastcore/issues/304 and was resolved.

Unfortunately the solution only fixed the issue for a specific endpoint in ghapi only, other endpoints like in this ticket https://github.com/fastai/ghapi/issues/157 or this one [fastai/ghapi/metadata.py#L1027](https://github.com/fastai/ghapi/blob/4fe9459c7e7bd631c7317eaa9d2365f14f2cb34d/ghapi/metadata.py#L1027) don't have a `archive_format` parameter in the url, and thus it ends up failing when trying to decode the response.

To me it seems like the initial solution suggested here https://github.com/fastai/fastcore/issues/304 of accepting the `decode` parameter in `urlsend` is a more generic solution.
This would not be the entire solution as some changes will still need to be implemented on the ghapi's side to actually send the `decode` argument when needed.

jph00:
I'm happy to look at a PR with a more general solution.

PR #611, opened 2024-08-19
add debug=True parameter to urlsend

radam9:
This PR is the first step to solving the issues related and similar to this https://github.com/fastai/ghapi/issues/157

To achieve this the parameter `decode` was added to the `urlsend` function to allow receiving non-JSON responses.

This fixes: https://github.com/fastai/fastcore/issues/608

### DIFF
--- a/fastcore/net.py
+++ b/fastcore/net.py
@@ -207,7 +207,7 @@ def summary(self:Request, skip=None)->dict:
     return res
 
 # %% ../nbs/03b_net.ipynb
-def urlsend(url, verb, headers=None, route=None, query=None, data=None, json_data=True,
+def urlsend(url, verb, headers=None, decode=True, route=None, query=None, data=None, json_data=True,
             return_json=True, return_headers=False, debug=None, timeout=None):
     "Send request with `urlrequest`, converting result to json if `return_json`"
     req = urlrequest(url, verb, headers, route=route, query=query, data=data, json_data=json_data)
@@ -216,7 +216,7 @@ def urlsend(url, verb, headers=None, route=None, query=None, data=None, json_dat
     if route and route.get('archive_format', None):
         return urlread(req, decode=False, return_json=False, return_headers=return_headers, timeout=timeout)
 
-    return urlread(req, return_json=return_json, return_headers=return_headers, timeout=timeout)
+    return urlread(req, decode=decode, return_json=return_json, return_headers=return_headers, timeout=timeout)
 
 # %% ../nbs/03b_net.ipynb
 def do_request(url, post=False, headers=None, **data):
--- a/nbs/03b_net.ipynb
+++ b/nbs/03b_net.ipynb
@@ -620,7 +620,7 @@
    "outputs": [],
    "source": [
     "#|export\n",
-    "def urlsend(url, verb, headers=None, route=None, query=None, data=None, json_data=True,\n",
+    "def urlsend(url, verb, headers=None, decode=True, route=None, query=None, data=None, json_data=True,\n",
     "            return_json=True, return_headers=False, debug=None, timeout=None):\n",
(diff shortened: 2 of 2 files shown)


radam9:
@jph00 I have created a PR as you suggested, 

I am not sure if you want to remove the if statement checking for `archive_format` in the `route`, since the next step would be to fix the urlsend invocation in ghapi and this part should end up being obsolete.

I am new to `nbdev` so do let me know if there is anything that needs to be change.

jph00:
Thanks, looks good! Sorry I don't follow the `archive_format` question - can you provide a link and more details?

radam9:
sure,

So it all started with this ticket https://github.com/fastai/ghapi/issues/22, this lead to the addition of the snippet below to the [urlsend function](https://github.com/fastai/fastcore/blob/2bdf551ecfc7c88f3656545734a273b1f80413b6/fastcore/net.py#L210)

```python
if route and route.get('archive_format', None):
        return urlread(req, decode=False, return_json=False, return_headers=return_headers, timeout=timeout)
```

Later on the same error started popping up again but for different routes https://github.com/fastai/ghapi/issues/157, what all the routes have in common is that they expect a non-json response. So we should not decode the response but rather return it as is, and that is why I am adding the change in this PR.

There are quite some Github API Endpoints that request non-json content as shown below, 

source: https://github.com/fastai/ghapi/blob/master/ghapi/metadata.py
- `/orgs/{org}/migrations/{migration_id}/archive`
- `/repos/{owner}/{repo}/actions/artifacts/{artifact_id}/{archive_format}`
- `/repos/{owner}/{repo}/tarball/{ref}`
- `/repos/{owner}/{repo}/zipball/{ref}`

as you can see, only one of the endpoints has `archive_format` in the route. 

After merging this PR I would like to work on adding logic in ghapi to properly set the `decode=False` when calling `urlsend` for all the appropriate endpoints, at which point the check for `archive_format` shown in the first snippet in this comment would be redundant.

So I am wondering if you think it would be better to remove that check now, or at a later point in time?

radam9:
> Thanks, looks good! Sorry I don't follow the `archive_format` question - can you provide a link and more details?

sure,

So it all started with this ticket https://github.com/fastai/ghapi/issues/22, this lead to the addition of the snippet below to the [urlsend function](https://github.com/fastai/fastcore/blob/2bdf551ecfc7c88f3656545734a273b1f80413b6/fastcore/net.py#L210)

```python
if route and route.get('archive_format', None):
        return urlread(req, decode=False, return_json=False, return_headers=return_headers, timeout=timeout)
```

Later on the same error started popping up again but for different routes https://github.com/fastai/ghapi/issues/157, what all the routes have in common is that they expect a non-json response. So we should not decode the response but rather return it as is, and that is why I am adding the change in this PR.

There are quite some Github API Endpoints that request non-json content as shown below, 

source: https://github.com/fastai/ghapi/blob/master/ghapi/metadata.py
- `/orgs/{org}/migrations/{migration_id}/archive`
- `/repos/{owner}/{repo}/actions/artifacts/{artifact_id}/{archive_format}`
- `/repos/{owner}/{repo}/tarball/{ref}`
- `/repos/{owner}/{repo}/zipball/{ref}`

as you can see, only one of the endpoints has `archive_format` in the route. 

After merging this PR I would like to work on adding logic in ghapi to properly set the `decode=False` when calling `urlsend` for all the appropriate endpoints, at which point the check for `archive_format` shown in the first snippet in this comment would be redundant.

So I am wondering if you think it would be better to remove that check now, or at a later point in time?

jph00:
Thank you for the very clear explanation! I think it's better to remove it later after those changes -- but if you feel differently after you start working on that, feel free to suggest otherwise, since I don't have a strong opinion.

PR #613, opened 2024-08-23
fix annotations generated by py2pyi

comhar:
Currently py2pyi generates fully qualified names (e.g. mod.submod.func) for annotations of user-defined types. These fully qualified names don't match the import statements in the node tree which can break downstream applications such as "jump to definition" in IDEs.

This PR resolves that issue by adding the correct import paths to the annotations.


comhar:
@jph00

jph00:
Nice job!

PR #621, opened 2024-09-11
add llm context files

hamelsmu:
@jph00 This adds the automation to generate `llms-ctx.txt`, `llms-ctx-full.txt` from the `llms.txt` file. 

I looked into how you are doing it in FastHTML, it IIUC you are manually executing it in a bash script.  I took it one tiny step further and made `update.sh` a pre-render script in Quarto.  

As far as the difference between `llms-ctx.txt` and `llms-ctx-full.txt`, I went with your intuition of having the filelist of symbols be available in the short version, but most of the docs available in the full version.  I am not sure what is going to work best here but I think it is worth trying it, potentially.

### DIFF
--- a/nbs/_quarto.yml
+++ b/nbs/_quarto.yml
@@ -1,5 +1,9 @@
 project:
   type: website
+  pre-render: 
+    - ./update.sh
+  resources: 
+    - "*.txt"
   preview:
     navigate: false
     port: 3000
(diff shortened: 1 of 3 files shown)


hamelsmu:
Ahh that sucks I thought I was being clever, but it doesn't really work because you need to publish the docs first

jph00:
I'm guessing we need to install https://github.com/AnswerDotAI/pysymbol-llm in the quarto workflow for this to work?

hamelsmu:
> I'm guessing we need to install https://github.com/AnswerDotAI/pysymbol-llm in the quarto workflow for this to work?

Good catch.  I guess I should add it as a dev dependency?

hamelsmu:
@jph00 Looks like I forgot to push the changes to pypi properly, but this is fixed now.  
I also fixed the newline and another spacing issue I noticed inside the backticks. 

See: https://fastcore.fast.ai/apilist.txt

I also went ahead and generated the llms-ctx.txt file (that is being deployed now)

PR #624, opened 2024-09-13
auto-update context

hamelsmu:
@jph00  This automates the creation of the context files. 

I am going to also update the llmstxt cli tool to also check for where the nbdev _docs directory is after this PR instead of hardcoding it at `/_docs` in that separate repo.

This PR depends on https://github.com/AnswerDotAI/llms-txt/pull/9

### DIFF
--- a/nbs/_quarto.yml
+++ b/nbs/_quarto.yml
@@ -1,7 +1,9 @@
 project:
   type: website
   pre-render: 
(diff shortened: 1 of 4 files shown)


hamelsmu:
Ahhh actually this doesn't work when I put in in GH Actions 😢 I need to update the CLI a bit to handle the `_docs` path better!

hamelsmu:
@jph00 ok we got it sorted!  This was one a bit confusing but finally got it!

PR #627, opened 2024-09-18
move nbdev detection to fastcore

hamelsmu:
This migrates nbdev detection to fastcore.  Once we review this PR we need to re-run the CI in this PR https://github.com/fastai/nbdev/pull/1449


fastcore/nbdev.py
```diff
@@ -0,0 +1,24 @@
...
+
+def nbdev_config_file(cfg_name=_nbdev_cfg_name, path=None):
+    "Get the nbdev config file path"
+    cfg_path = path = Path.cwd() if path is None else Path(path)
+    while cfg_path != cfg_path.parent and not (cfg_path/cfg_name).exists(): cfg_path = cfg_path.parent
+    if not (cfg_path/cfg_name).exists(): cfg_path = path
+    return cfg_path/cfg_name
+
+# %% ../nbs/13_nbdev.ipynb
+def is_nbdev():
```
hamelsmu, line 22: I would normally make this a one-liner, but I think having a a docstring is good for LLMs?

jph00:
I don't think it makes sense to create a new submodule just for this. Have a look at #628 and see if we can use that in llms-txt and nbdev instead?

PR #631, opened 2024-10-02
Nested attr supporting keys

Isaac-Flath:
Added support for keys in `nested_attr`

@jph00

### DIFF
--- a/fastcore/basics.py
+++ b/fastcore/basics.py
@@ -717,6 +717,7 @@ def nested_attr(o, attr, default=None):
         for a in attr.split("."):
             if hasattr(o, a): o = getattr(o, a)
             elif hasattr(o, '__getitem__'): o = o[a]
+            else: return default
     except (AttributeError, KeyError): return default
     return o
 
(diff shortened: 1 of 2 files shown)


jph00:
Can you look at the failing CI @Isaac-Flath ? Lemme know if you need help fixing it.

jph00:
That was fast :)

jph00:
Got another CI issue @Isaac-Flath !

Isaac-Flath:
Argh, another issue.  I am going to sleep on this and fix it in the morning.  I will get the fastai testing figured out for local tests so I can test it before the next attempt.

Isaac-Flath:
@jph00 All tests pass now, but it feels like there should be a better way.   I tried looking for a `__getattr__` method and tried checking for types, but wasn't able to get something that worked reliably in all cases without doing the `try` and catching the exception if it happens.

jph00:
@Isaac-Flath i think you want the hasattr() function

Isaac-Flath:
I used it in some of the previous versions that were failing the integration tests, and tried several things using it and kept failing test.  But just added it back and it seems to work on local fastcore and fastai tests, so I must have been doing something different wrong.

Anyway - much better.  Thank you!

jph00:
Thanks @Isaac-Flath . In your other PR that uses this, please update settings.ini to require this version of fastcore , and at-mention when that's ready, and I'll release both.

PR #635, opened 2024-10-08
Add support for type casting in the typed decorator

ncoop57:
This PR adds support for the `typed` decorator to perform automatic casting when the `cast` flag is set to `True`. It maintains the current API for backwards compatibility so that users can now specify the following:

```python
@typed
def discount(price:int, pct:float) -> float:
    return (1-pct) * price

with ExceptionExpected(TypeError): discount(100.0, .1)
# or

@typed(cast=True)
def discount(price:int, pct:float) -> float:
    return (1-pct) * price

assert 90.0 == discount(100.5, .1) # will auto cast the float "100.5" to the int "100"
```

The conversions are handled by a `type_map` dictionary that maps the expected type (i.e. the one specified as a type hint)
to the function to perform the conversion.

This allows for better validation handling for libraries like FastHTML instead of immediately throwing an error for common cases that would be expected to be properly handled like the str "on" to be convert to the int 1 or the str "none" to be converted to the int 0.

TODOs before merging:
- [x] Add better type conversion functions for the `type_map` that handles common cases such as converting the string "True" to the bool True if the expected type is a bool and the given one is a str.
- [x] Handle custom data types
- [x] Handle optional types (e.g., `a: int:None`
- [ ] ~~Handle container types like `list[str]` or `set[int]`~~ Not a priority. Can revisit if it is needed.


jph00:
lgtm @ncoop57 ! And yeah for the typemap values, just use the various `str2...` functions we have, moving them from fasthtml into here as needed (and updating fasthtml to use them, and updating its settings.ini fastcore dep). Sound OK?

ncoop57:
@jph00 Okay, I think the PR is ready for a full review!

jph00:
You got some conflicts @ncoop57 !

ncoop57:
@jph00 whoops! Should be good now 🤓

ncoop57:
@jph00 hold up on merging!

ncoop57:
Looking at an error when previewing the docs locally

ncoop57:
@jph00 okay, fixed the issue I had previewing the docs locally and made some slight tweaks by hiding some additional tests that I don't think need to be in the docs as it didn't look that great when previewing. Lmk what you think!

jph00:
lgtm!

PR #636, opened 2024-10-09
rm `progress=False` from tests of `parallel` since that is the default

hamelsmu:
It's the default so perhaps shouldn't be in the tests.

### DIFF
--- a/nbs/03a_parallel.ipynb
+++ b/nbs/03a_parallel.ipynb
@@ -141,6 +141,26 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
+   "outputs": [
+    {
+     "data": {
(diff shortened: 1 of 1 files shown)


nbs/03a_parallel.ipynb
```diff
@@ -286,15 +286,26 @@
       "text/markdown": [
```
hamelsmu, line 286: <p>Removed <code>progress=False</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/fastcore/pull/636/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

PR #645, opened 2024-11-03
Add support for Python 3.11-3.12

hugovk:
Python 3.11 was released in October 2022 and 3.12 in October 2023:

https://devguide.python.org/versions/

We can't add 3.13 yet (October 2024) because PyTorch doesn't fully support it yet:

https://github.com/pytorch/pytorch/issues/130249


jph00:
Thank you.

PR #657, opened 2024-12-13
remove unused var from parse_docstring

pratapvardhan:
This PR cleans up unused line of code – `docs = docstring(sym)`


pratapvardhan:
@jph00 since I can't add reviewers, cc'ing here – would need approval for workflow to run for this tiny cleanup.

jph00:
Thanks!

PR #659, opened 2024-12-22
Update Config to support ConfigParser keyword arguments

ncoop57:
This PR adds a `cfg_kwargs` argument to the Config object in the foundations module. This allows a user to pass keyword arguments directly to the ConfigParser object, allowing for things like overriding the behavior on how inline comments are handled. An example showcasing this functionality is also added to the docs.

### DIFF
--- a/nbs/02_foundation.ipynb
+++ b/nbs/02_foundation.ipynb
@@ -256,6 +256,14 @@
    "metadata": {},
    "outputs": [
     {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/nathan/.venv/lib/python3.12/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html\n",
+      "  from .autonotebook import tqdm as notebook_tqdm\n"
+     ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


ncoop57:
The integration test failure looks to be unrelated

jph00:
Yeah the integration test issue is due to a new bug in scipy, which has been fixed upstream but not released yet.

nbs/02_foundation.ipynb
jph00, line 1: Could you install ipywidgets (IIRC) and re-run the cell with `/Users/nathan/.venv/lib/python3.12/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html` so that warning doesn't turn up in the docs?

jph00 (changes requested):
Thank you! Just a minor issue below.

nbs/02_foundation.ipynb
ncoop57: Okay, this is fixed now, good catch!

ncoop57: @jph00 anything else needed to get this merged in?

jph00: Oops sorry got a bit behind over the xmas period! :O

PR #672, opened 2025-03-20
Fix colab link

galopyz:
Currently, the colab link, `colab_link('index')`, is set to index points to [index notebook](https://colab.research.google.com/github/fastai/fastcore/blob/master/nbs/index.ipynb), which is not the tour page. Therefore, the description, `if you want to follow along with this tour, click the link below:` is misleading.

This PR fixes this issue by changing to `"colab_link('000_tour')"`, which  generates  colab link to [tour notebook](https://colab.research.google.com/github/AnswerDotAI/fastcore/blob/master/nbs/000_tour.ipynb).

### DIFF
--- a/nbs/000_tour.ipynb
+++ b/nbs/000_tour.ipynb
@@ -48,7 +48,7 @@
     {
      "data": {
       "text/markdown": [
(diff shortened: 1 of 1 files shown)


jph00:
Thanks!

PR #674, opened 2025-03-22
Add a setter to patch

galopyz:
This PR adds a setter to `patch_to` and `patch` with `set_prop=True`.

Here are the changes:
- Creating a `setter` is limited to classes that already have a property with the same name. 
- It raises an `AttributeError` if there is no property with the name.


jph00 (changes requested):
Looks great! I suspect you can collapse your implementation down to one line if you follow my suggestion - I don't think we need all that extra ceremony, since python will already raise a suitable exception afaik.

jph00:
It looks like you might have forgotten to update the notebook with my suggestion? AFAICT it's only in the .py file, which means it'll be overwritten next time someone exports.

galopyz:
I am very sorry about that. I forgot to export indeed. I can fix that really quick.

jph00:
Thank you :)

PR #675, opened 2025-03-29
Add `headers` arg to `urljson`

CarloLepelaars:
Hi @jph00, I love the simplicity of `urljson` and use it often, but unfortunately it doesn't support `headers`.

Currently we can achieve the same behavior with `urlread`:

```python
urlread("https://example.com", headers={"API_KEY": key}, return_json=True)
```

but with the proposed change it becomes slightly more elegant

```python
urljson("https://example.com", headers={"API_KEY": key})
```


jph00:
Thank you! :)

PR #676, opened 2025-03-31
Add warning callout re: bool and store_true defaults

johnowhitaker:
Using bool arg with default True results in the default getting over-written (see e.g. https://github.com/AnswerDotAI/fastcore/issues/533) which has confused a few people. This PR adds a small warning to the docs to use `bool_arg` instead.

### DIFF
--- a/nbs/06_script.ipynb
+++ b/nbs/06_script.ipynb
@@ -96,6 +96,21 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "::: {.callout-warning}\n",
+    "## Boolean Arguments Default to False\n",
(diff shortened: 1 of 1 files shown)


johnowhitaker:
@jph00 it's a small thing but I've made this mistake in the past (setting `my_param:bool=True` not realizing it will still default to False). This PR adds a small callout to the docs with a warning to use `bool_arg` instead of `bool` in this case.
Demo of what I'm talking about: https://gist.github.com/johnowhitaker/52ae8cbeddaf43a62d345253f515ac97

jph00:
Thank you!

Issue #678, opened 2025-05-01
timed_cache appears to be FIFO instead of LRU


pydanny:
from https://nbsanity.com/static/c829aad0449a13ad7c6d63125e84f51d/timed_cache_bug.html

```python
from fastcore.xtras import timed_cache
from functools import lru_cache
from random import randint

@lru_cache(maxsize=2)
def lru_func(v): 
    return randint(1, 1000)

@timed_cache(maxsize=2)
def timed_func(v): 
    return randint(1, 1000)

assert lru_func(1) == lru_func(1)
assert lru_func(2) == lru_func(2)
assert lru_func(3) == lru_func(3)
assert lru_func(1) == lru_func(1)

assert timed_func(1) == timed_func(1)
assert timed_func(2) == timed_func(2)
assert timed_func(3) == timed_func(3)
assert timed_func(1) == timed_func(1)
```

The timed_func throws this exception:

```python
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[1], line 3
      1 assert timed_func(1) == timed_func(1)
      2 assert timed_func(2) == timed_func(2)
----> 3 assert timed_func(3) == timed_func(3)
      4 assert timed_func(1) == timed_func(1)

AssertionError: 
AssertionError: 
Traceback (most recent call last):

  File "/usr/local/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3667, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  File "<ipython-input-1-a6e3e0c59f0b>", line 3, in <module>
    assert timed_func(3) == timed_func(3)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

AssertionError
```

PR #679, opened 2025-05-02
Ensure flexicache is LRU cached

pydanny:
Originating ticket: #678

## Problem

The Python docs say that [dict.popitem is LIFO](https://docs.python.org/3/library/stdtypes.html#dict.popitem), which means it is **Last In First Out**. That is the opposite of FIFO, which is part of the algorithm often used in LRU (Least Recently Used).

Reference: https://docs.python.org/3/library/stdtypes.html#dict.popitem

## Solution

Since the cache is constantly being popped and reset, treat the order of the cache keys as an iterator and pick the first cache key set. To do this we just call the `next` command once on `iter(cache)`.

```python
if len(cache) > maxsize: del cache[next(iter(cache))]
```

## Tests

As cache invalidation is hard, new tests are added to flexicache that document cache invalidation and resetting.

## Alternative approach: OrderedDict

`collections.OrderedDict` has a `popitem()` method with a `last` boolean that replicates our design, as show in the code example below:

```python
cache = OrderedDict()
...
if len(cache) > maxsize: cache.popitem(last=False)
```

I rejected this approach because it adds an extra import and `cache.popitem()` is probably a very tiny bit slower than `del cache[next(iter(cache))]`. However, this approach is arguably more clear. In any case, I'm not opposed to changing out the current solution for the one presented here.

### DIFF
--- a/fastcore/xtras.py
+++ b/fastcore/xtras.py
@@ -748,8 +748,9 @@ def is_namedtuple(cls):
 def flexicache(*funcs, maxsize=128):
     "Like `lru_cache`, but customisable with policy `funcs`"
     import asyncio
+    from collections import OrderedDict
     def _f(func):
-        cache,states = {}, [None]*len(funcs)
+        cache,states = OrderedDict(), [None]*len(funcs)
         def _cache_logic(key, execute_func):
             if key in cache:
                 result,states = cache[key]
@@ -763,8 +764,7 @@ def flexicache(*funcs, maxsize=128):
                 cache[key] = cache.pop(key)
                 return result
             cache[key] = (newres, [f(None) for f in funcs])
-            # remove the oldest item when cache overflows
-            if len(cache) > maxsize: del cache[next(iter(cache))]
+            if len(cache) > maxsize: cache.popitem(last=False)
             return newres
 
         @wraps(func)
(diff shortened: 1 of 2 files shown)


jph00:
Good find @pydanny, thanks! Let's use the OrderedDict approach, since we value simplicity most highly. BTW looking at imports, I see that asyncio is imported in that function and not used, so perhaps replace it with the collections import?

jph00:
Oh BTW don't forget `nbdev_clean`. Otherwise the diffs are hard to read in PRs.

pydanny:
I switched the PR to use the OrderedDict approach. As for the use of `asyncio`, I moved it out of the decorator and to the initial module imports. It is used in `async_wrapper` of `flexicache`.

jph00:
Oh sorry you're right I didn't see asyncio being used at the bottom there! OK in that case you should put it back where it was -- in fastcore.xtras we prefer to only import rarely used modules if/when they're actually used. You should also only import `collections` inside the function for the same reason.

pydanny:
Why are all the tests now passing?

I was stumped by the failing test that had emerged, being unable to replicate it locally. Here is the failing command:

```sh
nbdev_test --flags '' --n_workers 3 --pause 1.5 --file_re "[0-9][0-1].*"
```

Now having moved the imports to within the function all the tests are now passing. Why is that the case?

jph00:
I don't know - I didn't see the failing tests so can't really guess. But glad it's working now :)

pydanny:
I still can't replicate it locally. For now I'm going to consider it a temporary GitHub problem.

PR #708, opened 2025-11-24
address darwin system parallel threading issues, without numpydoc fixes

s01st:


jph00:
Thanks for this. Can you run `nbdev_export` please, otherwise CI will fail when I try to merge.

I don't actually know what's happening in this PR. Can you tell me about the problem and solution?

s01st:
@jph00  thoughts? Or how else should I address this?

fastcore/parallel.py
```diff
@@ -30,7 +30,11 @@ def g(_obj_td, *args, **kwargs):
             _obj_td.result = res
         @wraps(f)
         def _f(*args, **kwargs):
-            res = (Thread,Process)[process](target=g, args=args, kwargs=kwargs)
+            if process:
```
jph00: I don't think we want this, do we? I'd expect `process=True` to give us a normal `Process`. Why would we want something different on Mac?

fastcore/parallel.py
```diff
@@ -158,7 +164,8 @@ async def limited_task(item):
 # %% ../nbs/03a_parallel.ipynb
 def run_procs(f, f_done, args):
     "Call `f` for each item in `args` in parallel, yielding `f_done`"
-    processes = L(args).map(Process, args=arg0, target=f)
+    Proc = get_context('fork').Process if sys.platform == 'darwin' else Process
```
jph00: `fork` feels a bit unexpected to me here too.

jph00 (commented):
I'm not sure about these changes, but certainly interested in discussing them. Would love to hear your more detailed thoughts.

s01st:
## Comments

### Regarding `Process`
> Disclaimer my understanding of this is not the best, but I know a bit about concurrency // distributed systems.
> 
> My understanding is that what Process does depends on default start method of the platform (which I suppose, could change in an update, but not likely).
> 
> Anyway, I believe for true linux systems Process forks a copy while macOS (apple silicon, maybe the old intel chips too) spawns a new interpreter.
> 
> So supposing threaded(process=True) linux ends up with the forked process and macOS does spawn which carries the implications of a fresh interpreter, reloading the module, and requires picklablity (which is its own headache in its own way)
> 
> There is a push to using spawn, but at the moment picklability + nested functions will likely lead to errors (which I've encountered)
> 
> So the below get_context('fork').Process is the special-case to make process=True behave the same way on macOS as it does on Linux — not to make it special, but to avoid macOS being the odd one out.
> 
> Provided that I understand.


### Regarding `OBJC_DISABLE_INITIALIZE_FORK_SAFETY`

> That certainly helps. Do we want to force that as fix? or UserWarning?

## Screenshots

Screenshot 2026-01-28 at 9 27 19 AM

Screenshot 2026-01-28 at 9 27 26 AM

Screenshot 2026-01-28 at 9 27 33 AM

jph00:
Funnily enough I actually hit this today when running a test on my mac, so just popped in a fix while I was there - I haven't checked, but I suspect it's probably basically the same as you had here. Take a look - if you think it's missing anything, lemme know.

PR #731, opened 2025-12-31
add slice support for L assignment

ncoop57:
This PR allows for `L` objects to now handle slice assignments:

```python
lst = L([1, 2, 3, 3, 4, 5])
lst[1:1] = [10, 20]
lst
# [1, 10, 20, 2, 3, 3, 4, 5]
```


jph00:
Thanks! FYI you can just say `if isinstance(idx, (int, slice))`. I'll make that change locally.

ncoop57:
🤯 that's so cool!

PR #741, opened 2026-01-12
Make handling of `~` consistent, along with error handling.

PiotrCzapla:
While using solveit, I noticed `create("~/memo.md")` was creating a file literally named `~` rather than expanding to the home directory. `view` already had `expanduser().resolve()`, but the other tools (`create`, `insert`, `str_replace`, `replace_lines`, `move_lines`) didn't.
Rather than adding `expanduser().resolve()` to each function individually, I've added a helper and wrapped `replace_lines`/`move_lines` in try/except for consistent error handling. Also switched from `if not ... return` to `assert`.

Let me know what you think.

### DIFF
--- a/nbs/12_tools.ipynb
+++ b/nbs/12_tools.ipynb
@@ -461,17 +461,17 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Directory contents of /app/data/repos/fastcore/nbs:\n",
-      "/app/data/repos/fastcore/nbs/000_tour.ipynb\n",
-      "/app/data/repos/fastcore/nbs/00_test.ipynb\n",
-      "/app/data/repos/fastcore/nbs/01_basics.ipynb\n",
-      "/app/data/repos/fastcore/nbs/02_foundation.ipynb\n",
-      "/app/data/repos/fastcore/nbs/03_xtras.ipynb\n"
+      "Directory contents of /path:\n",
+      "/path/000_tour.ipynb\n",
+      "/path/00_test.ipynb\n",
+      "/path/01_basics.ipynb\n",
+      "/path/02_foundation.ipynb\n",
+      "/path/03_xtras.ipynb\n"
      ]
     }
    ],
    "source": [
-    "print(view('.', (1,5)))"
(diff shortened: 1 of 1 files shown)


fastcore/tools.py
```diff
@@ -159,19 +164,21 @@ def move_lines(
...
+        assert not(start_line <= dest_line <= end_line + 1), "Destination within source range"
+        
+        chunk = lines[start_line-1:end_line]
+        del lines[start_line-1:end_line]
+        # Adjust dest if it was after the removed chunk
+        if dest_line > end_line: dest_line -= len(chunk)
+        lines[dest_line-1:dest_line-1] = chunk
+        p.write_text('\n'.join(lines) + '\n')
+        return f"Moved lines {start_line}-{end_line} to line {dest_line}"
+    except Exception as e: return f'Error: {str(e)}'
```
jph00: The assertion errors have sufficient info in their descriptions, but other errors perhaps need to type info too?
```suggestion
    except AssertionError as e: return f'Error: {str(e)}'
    except Exception as e: return f'Error: {repr(e)}'
```

fastcore/tools.py
```diff
@@ -144,12 +147,14 @@ def replace_lines(
...
-    p.write_text(''.join(content))
-    return f"Replaced lines {start_line} to {end_line}."
+    try:
+        p = valid_path(path)
+        content = p.readlines()
+        if not new_content.endswith('\n'): new_content+='\n'
+        content[start_line-1:end_line] = [new_content]
+        p.write_text(''.join(content))
+        return f"Replaced lines {start_line} to {end_line}."
+    except Exception as e: return f'Error replacing lines: {str(e)}'
```
jph00: This won't match the pattern "Error: " any more, which will break API clients relying on it. (Which I assume would be most clients - IIRC that's the pattern we use for indicating an error.)
```suggestion
    except Exception as e: return f'Error: {repr(e)}'
```

jph00 (changes requested):
Looks great. Just some minor suggestions.

jph00:
I'm gonna make those minor changes at my end, since I need this PR for what I'm doing now. :D

PR #744, opened 2026-01-15
Fix anno typing when `eval_str=True`

curtis-allan:
This PR fixes an issue with `docments` annotation type evaluation. 

When calling `docments(full=True, eval_str=True)`, empty annotation types in the output are not evaluated to `None` as they once were, causing issues in other functions that use `docments` for type setting (like `@call_parse`).

Testing was done using the `watch_export` cli, which would throw errors regarding `inspect._empty` not being a valid type previously. With this change it works as before


jph00:
Oh great fix! :)

PR #755, opened 2026-01-23
Add __for_llm__ auto-detection to llmtool

erikgaas:
Enhances `@llmtool` to auto-detect `__for_llm__` classmethod on the first param's class.

Adds helper functions:
- `params(f, exclude)` - get parameter annotations
- `extract_args(names, a, kw)` - extract args from positional/kwargs
- `call_via(via, method_name, a, kw, via_keys)` - create object via factory and call method
- `merge_hints(via, f)` - combine type hints from two functions

Usage:
```python
class Page:
    @classmethod
    def __for_llm__(cls, tab_id: int) -> Page: ...

@llmtool
def read(self:Page) -> str: ...
```

One decorator that patches the method AND creates a tool version.

### DIFF
--- a/fastcore/imports.py
+++ b/fastcore/imports.py
@@ -108,9 +108,45 @@ def is_usable_tool(func:callable):
 
 __llmtools__ = set()
 
-def llmtool(f):
-    assert is_usable_tool(f), f"Function {f.__name__} is not usable as a tool"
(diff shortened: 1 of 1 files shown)


jph00:
I guess this isn't that relevant any more

PR #769, opened 2026-02-02
fix urlretrieve NameError by importing tempfile

seemab-yamin:
Calling `urlretrieve` after `from fastcore.net import *` fails inside the function with

    NameError: name 'tempfile' is not defined

Root cause: `tempfile` is used but not imported in the module.

Changes:
- Added `import tempfile` to the top imports cell in `nbs/03b_net.ipynb`
- Ran `nbdev-export` → clean diff in `fastcore/net.py`
- Verified with local repro: download now succeeds without error

Tested with `nbdev-test` (passed).

### DIFF
--- a/fastcore/net.py
+++ b/fastcore/net.py
@@ -22,7 +22,7 @@ from .utils import *
 from .parallel import *
 
 from functools import wraps
-import json,urllib,contextlib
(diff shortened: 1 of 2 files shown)


jph00:
Good job!

PR #782, opened 2026-03-02
Add read_yml and write_yml Path extensions

motdam:
Adds Path.read_yml() and Path.write_yml() methods, following the same pattern as the existing read_json/write_json helpers.

PyYAML is imported lazily inside each function, with an ImportError message if it's not installed, so there's no new dependency added to the package.


jph00:
It's a good idea and implementation, but not something I want to add - thanks anyway! :)

PR #792, opened 2026-03-22
Add first and last methods to L

renato145:
Added helpers methods to L to get first and last elements given a filter function.


jph00:
Thanks! :)

PR #802, opened 2026-04-10
Add async mapping utilities: mapa and maybe_aiter

ncoop57:
Adds two new async utility functions:
 - `maybe_aiter`: Converts items to async generator if needed
 - `mapa`: Async version of map that handles both sync and async iterables/functions

 These utilities simplify working with mixed sync/async code by automatically handling both cases.


jph00:
Cool. I'll add a sleep(0) in that loop too so then we get the behavior we discussed the other day of not hogging the thread :)

PR #812, opened 2026-05-06
Lazy import `progress_bar`

KeremTurgutlu:
A lot of projects use `from fastcore.utils import *` to get the basic utilities, currently there is a huge bottleneck stemming from `fasthtml` -> `fastprogress` -> `fastcore.parallel`. This PR makes `progress_bar` a lazy import. Python shell scripts such as shell sage will benefit highly from this.

image


fastcore/parallel.py
```diff
@@ -118,13 +118,16 @@ def map(self, f, items, *args, timeout=None, chunksize=1, **kwargs):
         except Exception as e: self.on_exc(e)
 
 # %% ../nbs/03a_parallel.ipynb #f5327cb3
-try: from fastprogress import progress_bar
-except: progress_bar = None
+def import_progress_bar():
```
jph00, line 121: FYI I don't think this is worth having a function for - it replaces 2 simple lines of code used in 2 places, with 4 less-simple lines of code in the function plus 2 lines to call it! :)

jph00:
Thanks! Merging now - but follow up with a PR if you have a chance to remove the function. Or at least make it private.

PR #817, opened 2026-05-11
follow up #812

KeremTurgutlu:
Follow up on https://github.com/AnswerDotAI/fastcore/pull/812/changes#r3213959753


jph00:
I don't understand why this CI is failing. I'll try merging it anyway and see what happens.

jph00:
Ah @KeremTurgutlu this had a logic error:
```py
    try: from fastprogress import progress_bar
    except ImportError: return None
```
You can't `return None` there - should be `progress_bar=None`. I've actually removed this try/except now anyway.

KeremTurgutlu:
> Ah @KeremTurgutlu this had a logic error:
> 
> ```python
>     try: from fastprogress import progress_bar
>     except ImportError: return None
> ```
> 
> You can't `return None` there - should be `progress_bar=None`. I've actually removed this try/except now anyway.

Right! Sorry about that. Thanks!

PR #826, opened 2026-06-12
Fix patch_to to handle Enum and non-type iterables correctly

ncoop57:
## Problem

`patch_to` uses `tuplify(cls)` to iterate over the target class(es), but `tuplify` iterates over `Enum` members (and other iterables) rather than treating them as a single class, causing unexpected behaviour when patching Enum subclasses.

## Fix

Check if `cls` is a plain `type` first and wrap it in a tuple directly, falling back to `tuplify` only for sequences of classes. This ensures Enum classes (and any other iterable types) are treated as a single patch target.

Also fixes an unrelated CI failure: matplotlib 3.11 detaches a figure's canvas on `plt.close`, breaking `test_fig_exists` when run via nbdev's inline backend. It now re-attaches an Agg canvas if needed.

## Changes
- `fastcore/basics.py`: Replace bare `tuplify(cls)` with `(cls,) if isinstance(cls, type) else tuplify(cls)` in `patch_to`
- `fastcore/test.py` / `nbs/00_test.ipynb`: Make `test_fig_exists` re-attach a `FigureCanvasAgg` when the figure's canvas has been detached (matplotlib >= 3.11)


jph00:
Thank you :)

PR #835, opened 2026-07-02
Fix call_parse to reliably detect direct script execution

ncoop57:
## Summary

Replaces the fragile `inspect.getmodule()`-based `__main__` check in `call_parse` with a dedicated `_is_script_run()` helper that correctly identifies direct script execution across all three invocation styles: `python foo.py`, `python -m foo`, and `%run foo.py` in notebooks. Also fixes in-process nested runners (`nested=True`), broken since the introduction of `_in_call_parse`.

## Changes

- **Add `_is_script_run(frame)`**: resolves and compares `__file__` from globals against `frame.f_code.co_filename` to confirm the frame is the top-level body of a directly-run script — not an import or nested call.
- **Factor argv parsing out of `_f` into `_run_cli`**: auto-execution now calls it directly, since at that point we already know the file is a script entry point. Previously auto-exec re-entered `_f`, which infers what to do from its caller's frame — and on the auto-exec path that frame is `call_parse` itself, giving the wrong answer. This restores in-process nesting (a `nested=True` runner launching another script via `runpy`): the inner script parses the leftover `sys.argv` even while the outer runner holds `_in_call_parse`, which previously suppressed it.
- **`_f` itself is unchanged**: decorated functions calling each other still get plain calls (#759), and interactive zero-arg calls (notebook cells, the REPL) remain plain calls rather than argv parses.
- **Direct globals mutation**: replaces `setattr(mod, ...)` with `frame.f_globals[func.__name__] = _f` to avoid the `inspect.getmodule()` call that could misidentify the calling context.

## Motivation

The previous check could incorrectly fire (or fail to fire) when `call_parse`-decorated functions were imported into a `__main__` module, or called programmatically from a script. Additionally, nested runners silently passed defaults instead of parsing leftover `sys.argv`, since the inner script's entry point was treated as a plain function call.


jph00:
Thanks!

PR #846, opened 2026-07-09
Use string_classes constant instead of hardcoded (str, bytes) tuples

galopyz:
fastcore/imports.py already defines string_classes = (str, bytes), but L.flatten and urlopen use the hardcoded tuple directly. This PR updates those occurrences to use the named constant.


jph00:
Thanks!

PR #848, opened 2026-07-09
Fix mk_cell overwriting caller-provided outputs and execution_count

ncoop57:
## Summary

`mk_cell` was unconditionally clobbering `outputs` and `execution_count` for code cells, even when the caller explicitly passed values for them. It also defaulted
`execution_count` to `0` instead of `None` (the Jupyter convention for unexecuted cells).

## Changes

- Use `kwargs.setdefault('outputs', [])` and `kwargs.setdefault('execution_count', None)` so caller-supplied values are respected.
- Default `execution_count` is now `None` instead of `0`, matching standard Jupyter notebook and nbdev-clean behavior.


jph00:
Oh great thanks! :)

